PR #1, opened 2019-07-30
Some small updates to start contributing

DirkSonguer:
Added a README that explains what the project is about. This is based on our discussions on Twitter. It also contains some things I ran into and might be good to know to start.

I also added some missing assets to the StereoKitTest/Assets folder that were available in StereoKitCTest/Assets.


maluoi:
This is awesome! Thanks a bunch =D

I was been putting off documentation just because of how fast I expect things to change, this will totally be helpful regardless of changes!

Issue #3, opened 2019-07-31
[CS] There is no way of knowing if a correct xr_session has been created


DirkSonguer:
**Issue**: With C# it's not possible to know if the OpenXR initialization has succeeded. The application will just run silently.

Steps to reproduce:
- Don't attach a Windows MR headset
- Run the StereoKitTest application

**Result**: The application starts normally, however nothing happens. As a developer I have no way of knowing if the application has a correct OpenXR xr_session running.

**Expected**: There is either an exception if no xr_session could be created or some other way of knowing.

**Details**: With CPP it is possible to know if the initialization has failed as sk_init() returns a bool. With C# there is no such thing as the constructor is just calling sk_init() and not listening to any return.

**Suggestion**: Change StereoKit class constructor to:
``` csharp
        public StereoKitApp(string name, Runtime runtime)
        {
            if (!sk_init(name, runtime))
            {
                throw new Exception("StereoKit: Could not initialize " + runtime.ToString() + " runtime");
            }
        }
```

**Proposal**: Add exceptions to StereoKitC library to let them properly bubble up, regardless of CPP or C#. At least for the fatal ones (like creating xr_session, events, actions, rendering etc.)

maluoi:
Something is definitely needed here! I remember writing that example and thinking, 'Huh, I can't check errors out of a constructor, can I?'. Then I moved on. There's definitely a lot of error checking I've deferred so far, it would be really nice to have a good logging system too.

I'm not a fan of catching exceptions, it falls into the realm of 'Magic! Where Did That Functionality Come From?' to me. So I generally try to avoid throwing them unless I actually intend to crash an application. I much prefer error codes!

So I think the core trouble here is that the constructor is doing a little too much work! I think my preference would be either a static StereoKitApp.Create() that can return null, or an instance member StereoKitApp.Initialize() that can return an error code. That and better logging :)

DirkSonguer:
Errors out of constructors are Exceptions 😅

Exceptions are only magic if you use them only once or twice and they come as a surprise. If you are using them throughout then they can be expected and people should code for it. As for error codes: This is what exception types are for.

If you want to avoid them at any cost, then I agree that having a separate StereoKitApp.Init() function would better mirror the C behavior:

``` csharp
        public StereoKitApp()
        {
                // This would be empty now? 
        }

        public Init(string name, Runtime runtime)
        {
            return sk_init(name, runtime);
        }
```

[...]

PR #15, opened 2019-08-02
Fix for "There is no way of knowing if a correct xr_session has been created"

DirkSonguer:
Fix for Issue #3 (There is no way of knowing if a correct xr_session has been created)

- Added initialization in line with the C behavior
- Changed C# example accordingly

Potential further improvements: Currently chaining the Initialization() calls fails (for example calling it once with Runtime.MixedReality and if that fails with Runtime.Flatscreen) because of some internal configuration that already happened. Ideally this should work.

### DIFF
--- a/StereoKit/StereoKit.cs
+++ b/StereoKit/StereoKit.cs
@@ -25,14 +25,18 @@ namespace StereoKit
         static extern float sk_timef();
         #endregion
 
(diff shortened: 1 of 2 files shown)


maluoi:
Looks great! I'll add the note as a bug.

PR #39, opened 2020-01-13
Fixed text_make_style C# binding returning void

sjando:
Fixed incorrect return type in the Text NativeAPI wrapper preventing TextStyle creation

(Really nice library btw!)


maluoi:
Nice catch! Thanks for the fix :)

Issue #46, opened 2020-02-22
ISpatialAudioClient for audio on WMR platforms


maluoi:
ISpatialAudioClient is the recommended method for rendering audio on a HoloLens! It does spatial audio and mixing on the HPU, so is largely free to use as far as the app is concerned. Win-win all around.

Reference implementation is here:
https://github.com/microsoft/spatialaudio-unity/tree/develop/Source/Spatializer/HoloLens2

MikeChemi:
I'd love to help out here :)

PR #64, opened 2020-03-05
Add ISpatialAudioClient support for Windows platforms

MikeChemi:
On HoloLens 2, spatial audio is offloaded to the HPU by using the ISpatialAudioClient API. This API is available on all version of Windows and provides the same sonic quality, so it makes sense to use this API whenever it is available. This PR enables spatial audio playback via the ISpatialAudioClient API (shortened to ISAC in the code). If ISAC is not available, it falls back to the original miniaudio implementation.

On desktop PCs, the use of ISAC requires the PC user to set their audio settings properly. I added a blurb about this in the documentation for the Sound class. If ISAC is not configured properly, audio playback falls back to the original miniaudio implementation. On HoloLens 2, there are no setup steps required.

### DIFF
--- a/StereoKitC/asset_types/sound.cpp
+++ b/StereoKitC/asset_types/sound.cpp
@@ -14,6 +14,7 @@ namespace sk {
 
 sound_inst_t au_active_sounds[8];
 float        au_mix_temp[4096];
+matrix       au_head_transform;
 
 ma_decoder_config au_decoder_config = {};
 ma_device_config  au_config = {};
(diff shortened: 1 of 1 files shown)


maluoi:
Woooh! Very exciting! I'm gonna put this through some paces and see how it goes :)

MikeChemi:
@maluoi please let me know if you hit anything weird. I did all the testing I could think of, but as this is my first foray into the code base I wouldn't be surprised if I missed anything

MikeChemi:
@maluoi ah yes, thank you for the head rotation. I didn't know how to get at it T_T In general, yes 1/r is all it takes for audio. The "weird thing" here is that if you are closer than 1m, the sound is supposed to get louder. However, the ISAC API has a hard cap for volume at 1.0f. This means that any sounds closer than 1m to you will all be the same loudness. As you get further away, you should start to hear the effect.

With all that in mind, is this OK to merge? I like to squash merge to keep things tidy.

maluoi:
Awesome, that wraps #46 !

PR #72, opened 2020-04-02
Fix the WINAPI_FAMILY_PARTITION and asset path

gongminmin:
Fix some minor issues:
1. winapifamily.h was not included, so WINAPI_FAMILY_PARTITION is not defined.
2. The start directory of StereoKitCTest project is in Examples/StereoKitCTest. It can't locate ../../../Examples/Assets.
3. Native projects' <WindowsTargetPlatformVersion> is not consistent with other projects.


maluoi:
Also I'm really curious, are you using StereoKit from the C side right now? If so, what's your reasoning there over C#?

gongminmin:
No, I just open the sln and compile the project, these problems jump out. The WINAPI_FAMILY_PARTITION one is a compiling warning. I have 20H1 SDK installed, choosing 10.0 in project file will pick that SDK. However, some interface of cppwinrt is changed, so it can't compile.

maluoi:
Sounds reasonable to me! Thank you for the fix! :)

Issue #84, opened 2020-05-30
OpenXR 1.0.9 Update


maluoi:
Lots of goodies: https://github.com/KhronosGroup/OpenXR-SDK/commit/97cfe495bb7a3853266b646d1c79e169387f9c7a

- [x] Extension XR_MSFT_spatial_graph_bridge
- [x] Extension XR_EXT_hand_tracking
- [x] Extension XR_MSFT_secondary_view_configuration

maluoi:
Done in 4ba510b2eacbc4ff754f50f631c7a82a9705d9c2, but still needs testing when the runtime supports the new extensions.

PR #85, opened 2020-06-16
Correctly initialise structures before locating hands with OpenXR

rblenkinsopp:
Correctly initialise the `XrHandJointLocationsEXT` structure members `jointCount` and `jointLocations` before calling `xrLocateHandJointsEXT`. Without this the call will fail with an incorrectly zero `jointCount`, or segfault accessing the null pointer for `jointLocations`.

This fixes support for XR_EXT_hand_tracking in #84


maluoi:
Thank you for this! Not a ton of examples for the hand tracking ext yet :)

The runtime was in a bit of an odd spot between preview and ext when I dropped that code in, so I wasn't able to test it fully! Looks like it's settled down now, so I'll check the rest of this over in the near future too!

PR #87, opened 2020-07-14
[hotfix] Switched back to replace method as replaceAll is not widely supported.

neogeek:
This PR reverts a recent change (030bc671124e014dba0e1edf5c080a6f92f302cd) wherein the method `replace` was swapped for `replaceAll`. Unfortunately, `replaceAll` is only supported in the latest version of Firefox and Safari. Both Chrome and Edge are currently missing this feature.

https://caniuse.com/#search=replaceAll

I've also added a global flag to the regular expression, replicating the functionality of the `replaceAll` method.

### DIFF
--- a/docs/src/docs.js
+++ b/docs/src/docs.js
@@ -84,14 +84,14 @@ function renderFolder(folder, parent, hierarchy) {
 
     // Link to the actual document
     var link = document.createElement("a");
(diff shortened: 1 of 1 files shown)


maluoi:
Nice catch, thanks for the fix! Firefox is my daily driver, I had no idea that was a new function!

PR #88, opened 2020-07-23
[feat] Renderer.Projection and Matrix Multiplication

neogeek:
This PR exposes the projection matrix from the renderer and adds the ability to multiply a `Matrix` by a `Vec4`.

These changes are to support adding features similar to the [Camera.WorldToScreenPoint](https://docs.unity3d.com/ScriptReference/Camera.WorldToScreenPoint.html) and [Camera.WorldToViewportPoint](https://docs.unity3d.com/ScriptReference/Camera.WorldToViewportPoint.html) in Unity.

### DIFF
--- a/StereoKit/Native/NativeAPI.cs
+++ b/StereoKit/Native/NativeAPI.cs
@@ -262,6 +262,7 @@ namespace StereoKit
 
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   render_set_clip      (float near_plane, float far_plane);
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   render_set_fov       (float field_of_view_degrees);
+		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern Matrix render_get_projection();
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern Matrix render_get_cam_root  ();
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   render_set_cam_root  (Matrix cam_root);
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   render_set_skytex    (IntPtr sky_texture);
--- a/StereoKit/Systems/Renderer.cs
+++ b/StereoKit/Systems/Renderer.cs
@@ -54,6 +54,8 @@ namespace StereoKit
 			set => NativeAPI.render_set_cam_root(value);
 		}
 
+		public static Matrix Projection => NativeAPI.render_get_projection();
+
 		/// <summary>Adds a mesh to the render queue for this frame! If the Hierarchy has a transform on it,
 		/// that transform is combined with the Matrix provided here.</summary>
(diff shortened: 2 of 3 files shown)


maluoi:
Heyas! I love the code, thanks :) Some things I'd like to touch on before adding this one in though!

`render_get_projection` is only valid during Flatscreen applications, and while I have surfaced a number of options like this so far, I'm generally a little hesitant to do so and need a better strategy to manage it! I'm trying to be careful about providing functions that only work in one and not the other!

So with that in mind, I'm curious what you're trying to accomplish with it? I imagine it's something like a picking ray? I do have a function in C called `ray_from_mouse` which has much more Flatscreenish terminology, and would provide fewer expectations of working in VR. It needs surfaced into C#, and perhaps a different location/association, but would that solve the issue for you as well?

neogeek:
Ah, this would explain why things are acting differently in the headset and on desktop.

I'm building an application that captures position data from a moving target and eye gaze data from the HoloLens 2. I need to convert the position of the target and the position of the raycast hit point from world space to screen space so I can render them on a 2D plot.

Thank you for the quick reply!

maluoi:
Ok! So you would definitely want access to this stuff in Stereo then.

I don't believe I'm actually storing OpenXR's projection and view matrices yet, so that'll definitely need to go in! I suppose it would be a function like Unity's, `WorldToScreenPoint` where you specify which eye in addition to the point.

With Flatscreen support, I'd use the exact same function, but disregard the eye provided, and return the same answer for each eye?

If you want to work on it, that'd be cool :) If not, I can get that in pretty quickly too.

maluoi:
Oh yeah, there's another tricky thing about stereo camera info and screen projection here too! 

MR uses the concept of predicted time plus re-projection to get the visuals in the right place. Right before render you predict where the headset will be by frame display time, and then the system re-projects the result to match up with where the headset actually ended up.

While I imagine the projection matrix doesn't change much through this, the eye viewpoint _definitely_ does. This probably makes it near impossible to get a practical result for WorldToScreenPoint for most purposes. Not to mention eye gaze isn't guaranteed to be on-screen. In VR, there's also a lot of distortion of the final image that happens to correct for optics, so getting a viewport position wouldn't even be all that practical for making a plot! I'm uncertain if that's true for AR as well, but I wouldn't be entirely surprised.

Maybe it would even be better for you to just raycast at a plane in front of the user's face? Or use the angle between the head's forward direction and the eye direction? I guess it depends on what you're trying to measure.

neogeek:
I'll take a look at OpenXR and its implementation of projection matrixes. Although I do have to admit, I am very out of my element when dealing with math in a 3d environment. Would it be ok if I floated some questions by you as I work on this? I am on the Microsoft HoloLens Slack if that is an easier form of communication.

We are already using the gaze pose to raycast to a plane at a fixed distance from the camera. Hopefully, that will help simplify things, at least in my implementation.

maluoi:
Yeah, slack is great! I'm usually on there, and I get notifications from it :)

And if the ray->plane method is working well enough for you, I think I'd be happy to skip on surfacing the projection matrix for now!

PR #102, opened 2021-01-20
- Bug fixed in mesh_gen_cylinder function

pambros:
Hi, just a small bug fixed in the mesh_gen_cylinder where the radius is used instead of the depth.


maluoi:
Thanks a bunch! You're totally right :)

What's wild is that since the normals are all the same anyhow, the surface looks completely flat!

PR #157, opened 2021-06-07
Linux EGL support (initial)

technobaboo:
StereoKit currently uses GLX for all its rendering on Linux, however as this is incompatible with SteamVR (texture arrays don't work with its OpenXR implementation) and EGL is much more portable, lightweight, easy to deal with, etc. it makes sense. And it's needed for Stardust and all :p


maluoi (commented):
This (dynamic linking of OpenXR) unfortunately is something I want to avoid doing, because it affects end-users of C# StereoKit apps. An end-user should not need to build OpenXR.

Alternatively, we could ship compiled openxr_loader with Linux builds, like Android requires, but this does increase file size, which is currently why I use static linking.

maluoi (commented):
We chatted about the details of this on Discord, but I'll add it here too just for the record. StereoKit still needs to be capable of working with SteamVR on Linux, which means GLX needs to stick around until a Vulkan backend can be made, or SteamVR can be initialized with EGL.

maluoi (commented):
FcConfig stuff looks good, if you want to do that as a separate PR, I'll take it right away.

technobaboo:
I know we did. It's not done yet, but also SteamVR doesn't work with Xlib
either...

technobaboo:
sure, how would I split it off that branch though?

maluoi:
> 
> I know we did. It's not done yet, but also SteamVR doesn't work with Xlib either...
> 

SteamVR should work right now if I use the same fallback as Android does. It should be a 2-3 line change. If I had SteamVR set up for testing on Linux, or if I knew how long it would take for Steam to get around to fixing the issue, I'd probably have done it already. I probably should take the time to do that soon though.

>
> sure, how would I split it off that branch though?
>

I... don't exactly know! I could probably do it with Fork, but I don't know the matter quite well enough to explain it.

slitcch:
FWIW, I made a slightly cleaned up version here: https://github.com/technobaboo/StereoKit/tree/feature/linux-EGL , still has the same no-Xlib problem and a crash on my machine, but that branch is all on-topic

slitcch:
No, nvm, I think Nova force-pushed some weird things to that branch

technobaboo:
I'll submit a new pull request with the branch that features GLX and EGL both working, separated by a compilation option.

PR #179, opened 2021-07-29
Allow disabling the physics system

technobaboo:
For most people, this isn't necessary... but given it takes 3/4 of the time compiling react physics I figured it was worth having the option. PR #176 should be applied before this one if I understand correctly.


maluoi:
I don't think I really love the way this is implemented? I'd prefer to see this approached as like a "passthrough physics backend" option, rather than erroring out when using physics APIs.

So for example, `demo_basics.cpp` and `input_hand.cpp` should still compile and run as they were, they just wouldn't do anything interesting.

PR #194, opened 2021-09-25
Add sliders for f64

slitcch:
Be careful merging this, the explosion of hslider variants you might get by going down this path sounds unpleasant, and I'm not totally sure I did everything right

### DIFF
--- a/StereoKitC/stereokit_ui.cpp
+++ b/StereoKitC/stereokit_ui.cpp
@@ -1521,8 +1521,8 @@ bool32_t ui_input_16(const char16_t *id, char16_t *buffer, int32_t buffer_size, 
 
 ///////////////////////////////////////////
 
-template<typename C>
-bool32_t ui_hslider_at_g(const C *id_text, float &value, float min, float max, float step, vec3 window_relative_pos, vec2 size, ui_confirm_ confirm_method) {
+template<typename C, typename N>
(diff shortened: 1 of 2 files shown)


StereoKitC/stereokit_ui.cpp
```diff
@@ -1592,7 +1592,8 @@ bool32_t ui_hslider_at_g(const C *id_text, float &value, float min, float max, f
 	}
 
 	if (button_state & button_state_active) {
-		float new_val = min + fminf(1, fmaxf(0, ((window_relative_pos.x-button_size.x/2)-finger_x) / (size.x-button_size.x)))*(max-min);
+		float pos_in_slider = fmin(1, fmax(0, ((window_relative_pos.x-button_size.x/2)-finger_x) / (size.x-button_size.x)));
+		N new_val = (N)min + (N)pos_in_slider*(N)(max-min);
 		if (step != 0) {
 			new_val = min + ((int)(((new_val - min) / step) + 0.5f)) * step;
```
maluoi, line 1598: This `0.5f` is technically just for rounding, but to reduce unit conversion back and forth in this case, it should be better to say `(N)0.5` or something.

slitcch: Oh indeed!

PR #197, opened 2021-10-13
Fix compilation failure for ARM64

maxgolov:
Hello, this is my first commit here. Apologies if it's been already resolved somewhere else.

I'm using latest Visual Studio 2019 (16.11.3) , targeting ARM64. Compilation of `render_sort.cpp` fails.

The fix is to check for both macros `_M_ARM` and `_M_ARM64`. After that everything compiles successfully.

Thank you.


maluoi:
That's interesting! I have been building this for ARM64 for a while, haven't run into this as an issue before. This does look more correct though, so I'll be happy to take it :)

I did switch this to `/develop` FYI, master branch is always the most recently published version.

Issue #198, opened 2021-10-14
Spatial graph bridge error checking and dynamic node support


sjando:
_World.FromSpatialNode_ works nicely in many cases but has a couple of limitations:

- The try pattern overload (via _world_try_from_spatial_graph_) does not correctly report failure in all cases. While the result of _xrCreateSpatialGraphNodeSpaceMSFT_ is checked, this call rarely if ever fails, even with a garbage/empty/random node GUID. The subsequent _openxr_get_space_ call will however fail in these cases, but that result is not checked. So the function returns true while the output pose is left unset.
- Both overloads are hardcoded in StereoKitC to use _XR_SPATIAL_GRAPH_NODE_TYPE_STATIC_MSFT_. It would be nice to also have a way to resolve the world pose for XR_SPATIAL_GRAPH_NODE_TYPE_DYNAMIC_MSFT nodes, which currently seem to fail silently as per the above point. As an example, the sensor rig node provided by the HL2 Research Mode API is a dynamic node that doesn't appear to work with the current implementation.

If there's a preferred way to expose a static vs dynamic option I'd be happy to take a crack at updating my local fix and submitting a PR.

maluoi:
I would kinda expect the sensor API to provide a timestamp in cases where that would make the most sense? Like a camera should report the picture's time so the pose of the camera can be correctly identified. 16ms of time difference can be a lot for some of those sensors, and something like a picture might come in much later than 16ms! If it's not giving a timestamp to you, I'm totally okay with ignoring that until it becomes an issue down the road :)

You may be right about the usage pattern there too, looks like it's also missing a call to xrDestroySpace. Not sure how costly that is in this case. I think I really wanted to wrap all this stuff up into a high-level concept of an Anchor, but it's a big topic that I haven't had a chance to tackle yet! That would absolutely help with managing the surrounding assets.

I have definitely been considering exposing some access to OpenXR for developers! It absolutely could make things easier for edge-cases. The trick is doing that in a portable way that doesn't inconvenience more casual users. For example, I wouldn't want the core StereoKit NuGet to require a dependency on OpenXR bindings. I think this may be doable in a reasonably nice way on the C++ side, but this idea does become trickier up in C#. Perhaps as an additional NuGet package, or some library exports that aren't included in the main package, but are documented as a sample?

sjando:
[...]

PR #200, opened 2021-10-17
Dynamic spatial node support

sjando:
Adds the ability to specify a spatial node type (static or dynamic) when calling World.FromSpatialNode, as well as the ability to specify a timestamp (Windows QPC ticks, internally converted to XrTime) at which to locate the spatial node. Default behaviour remains unchanged (static nodes using 'current' SK time). Also adds some extra checking to better handle the common failure case where a _xrCreateSpatialGraphNodeSpaceMSFT_ call succeeds but the subsequent _openxr_get_space_ call fails.

This enables the use of the World.FramSpatialNode API with sensor APIs including Windows.Media.Capture and Research Mode, which may utilise dynamic nodes and/or timestamped (typically QPC-based) data.

Addresses #198


maluoi:
Cool stuff, thanks for the submission! I'll test this out on Monday, but two bits stand out while reading through it:
- Your `XrTime time = 0;` definition is within a platform specific ifdef, but used outside that ifdef. I believe that probably will cause errors when compiling for other platforms :)
- Your C# binding for `bool32_t` is a `bool`, which is a trap that springs inconsistently on different platforms! It's safest to marshal as an `int` instead.

PR #258, opened 2021-12-13
Added the ability to get all Parameters on materials and also Get the Info of Parameters by index. 

Faolan-Rad:


maluoi:
Thanks for the PR! I'm out traveling, so I can't really test this out until Friday, and I may have a few additional thoughts later.

Some items that stand out in my mind though:
- `Param` vs `Parameter`: the existing API uses `Param` consistently, so `Param` should currently be the correct choice for name.
- Perhaps a bit nitpicky, but I like to avoid the use of `var` in core code.
- You may be mixing tabs and spaces? Alignment looks a bit off, SK's codebase uses tabs.
- I need to read up on tuples or anonymous structs, or however those may be called, and figure out some details about their implementation before I take this. SK's API currently does not use that syntax feature anywhere, and relies on `out` parameters instead. I guess that doesn't really jive in the context of the enumerator though, and I do like the way your enumerator looks. This may also have some ramifications on the documentation tool, which isn't likely to parse this function signature correctly.

Faolan-Rad:
Would you prefer me to use a struct instead of a tuple
``` csharp
		/// <summary>
		/// The name and type of a Parameter on a Material
		/// </summary>
		public struct ParamInfo
		{
			/// <summary>
			/// The name of the param
			/// </summary>
			public string Name;

			/// <summary>
			/// The type of the param 
			/// </summary>
			public MaterialParam Type;
			/// <summary>
			/// Parameter Info constructor 
			/// </summary>
			/// <param name="name">Name of the shader parameter.</param>
			/// <param name="type">New type for the parameter.</param>
			internal ParamInfo(string name, MaterialParam type)
			{
				Name = name;
				Type = type;
			}
		}
```

maluoi:
Yeah, that'll also totally work :)

Also, didn't notice this before, but if you could swap this to the `/develop` branch, that'd be excellent! `/master` always points to the latest stable NuGet release. I'll merge it after that!

PR #269, opened 2021-12-25
All Swizzels

Faolan-Rad:


maluoi:
This does seem a bit more excessive than I was thinking. Vec4 has like... 1024 permutations there?

Swizzles don't normally include 0, perhaps that can be dropped? That'd bring Vec4 back to 256, which is a slightly more reasonable number. Starting swizzles with O also introduces an intellisense issue I don't think I'd want in there either. Swizzle spam that starts with WXYZ groups all the way at the end, and doesn't really interfere with the other members, but O puts a bunch of that spam in the middle.

Faolan-Rad:
I could drop swizzles starting with zero and also remove the pointless ones like xyzw

Faolan-Rad:
Or could turn 0 into _

Faolan-Rad:
Or could remove all zeros

Faolan-Rad:
Have changed swizzles to be a separate static class so people need a using declare of StereoKit.Swizzels to use it

Faolan-Rad:
Was debating on adding swizzles to expand a float up

Faolan-Rad:
could also defer the using types to be a bit more specific as well an example is using StereoKit.Vec2Swizzels

maluoi:
Ok, so I've slept on this a bit, and I have a few more thoughts.

1. Adding the entire permutation of swizzles to C# feels like an attempt to make the language something that it's not. Kinda like writing macros in C to make the language syntax different. This is generally something I'm wary of, and I typically prefer idiomatic code.
2. Swizzles may end up getting used as a substitute for more descriptive code, like your earlier example, where `v.ZYX` could be used as shorthand to rotate 90 degrees clockwise around the Y axis. A more appropriate piece of code would be `Vec3.RotateY90CW(v)`, which is _much_ more clear about what it's doing.
3. Code is not zero cost, and a large number of permutations will increase binary size. Probably not that much in this case, but this is something I track and optimize for.
4. The properties that SK does have in the existing API are primarily about converting from one vector type to another, rather than swizzling values around. So while this may be a debatable or tangential point, I'm inclined to say SK doesn't actually have swizzles to begin with.

So with that all that, I don't think swizzles belong in StereoKit's core. They can still be handy to have around though, especially when trying to port complex shaders to the CPU, so I'd propose [putting them in the Tools folder instead](https://github.com/maluoi/StereoKit/tree/develop/Examples/StereoKitTest/Tools)? That way people can opt into them if they really want them, but they don't clutter anything up.

The best format for this would be a single C# file with all the swizzles that people can copy/paste, or right-click -> save to their own project. I did also like `_` instead of `0`, and would point out there's a few misspelled occasions of `Swizzels` that would need fixed too.

Faolan-Rad:
Moved Swizzles to one folder in the tools folder

PR #270, opened 2021-12-28
Virtual Keyboard

Faolan-Rad:
I added Virtual Keyboard with a couple of other things for the Virtual Keyboard like PushKeyboardFocusSteal and PopKeyboardFocusSteal


Faolan-Rad:
Forgot to to remove elements from stack when it is supposed to for PushKeyboardFocusSteal and PopKeyboardFocusSteal

maluoi:
This looks pretty neat! I've been wanting to have something like this for a while, I'm glad you've gotten to it here with layouts and everything :)

One thing to note first: I believe a virtual keyboard like this is probably a placeholder in the long run? Keyboards are an accessibility issue, and are most appropriately handled by the OS, like UWP does. I expect that over time, other MR platforms will also add their own MR keyboards that fully support language, speech, emoji, and probably a lot more. A feature like this should probably only be considered a fallback system.

---

You may have somewhat missed this, but SK does already have some virtual keyboard support! In particular, UWP provides an OS level MR compatible virtual keyboard that will pop up on HoloLens and WMR Desktop. The API for controlling this is in `platform_utils.h`, which has these functions:
```cpp
bool  platform_keyboard_available();
void  platform_keyboard_show     (bool visible);
bool  platform_keyboard_visible  ();
```

Ideally, SK's virtual keyboard should be a transparent extension of the OS functionality, where SK prefers OS provided MR keyboards when present, (or physical keyboards in simulator mode) and falls back to this keyboard system if no better solution is available. The developer shouldn't need to worry about keyboard layout or specific virtual keyboards implementations directly. This means that there probably shouldn't be much of a public API, though `platform_keyboard_show` and `platform_keyboard_visible` could be good candidates for the public API.

`ui_box_interaction_1h_pinch/poke` are currently the building blocks for interaction in the UI system, it's possible you may want to move some of the `nokeyboard_steal_stack_ids.add(id);` calls to those functions? Not 100% sure, but this may save from adding the code to each element.

**Here are some things I'd love to see changed:**
- Virtual keyboard layout should be automatically chosen based on the platform's language preference, `virtualkeyboard_setKeyboard_Layout` and family should not be part of the public API.
- Virtual keyboard's interface should go through the existing `platform_keyboard_` functions, rather than providing their own, so for example, `platform_keyboard_show` should call `virtualkeyboard_setopen` if no OS virtual keyboard is present, this is probably the only place `virtualkeyboard_setopen` should be called from.
- Since this virtual keyboard is a platform fallback, its lifecycle should be similar to other platform fallbacks. [Check out file_picker](https://github.com/maluoi/StereoKit/tree/master/StereoKitC/tools) for the existing example of this.
- Your commentdocs need a bit of love. If the commentdoc says the same thing as the variable/method name, there's no reason to have it. A good commentdoc should include units, details, and additional insight not available otherwise :)
- C code should be consistently snake_cased, and open brackets always go at the end of line.

Faolan-Rad:
Part of the reason why I added it to the public API is so people can set it to a Numpad if they want I guess would be a better solution would be to add an option to platform_keyboard to change the type of keyboard that is in use similar to the InputType on android [ref](https://developer.android.com/reference/android/text/InputType)

Faolan-Rad:
Currently does not switch to any other langs but can once the keyboard layout is set up and a way of finding what lang the user is in is added.

maluoi:
Looking better!

`UI.Input` should not take `useNativeKeyboard` as a parameter. This should be auto-determined by StereoKit based on availability, basically the behavior you have for passing true for that parameter. I'm not convinced that a dev should be allowed to force the fallback keyboard over the OS standard keyboard, but even if so, it should probably be a `platform_keyboard_` function instead of a `UI.Input` parameter.

The keyboard layout category issue presented with `KeyboardInputType` seems a lot more complicated now that I dig into the details. It looks like Android and UWP both don't like the idea of surfacing the layout directly, but instead give you text categories that translate into a mix of keyboard layout and text validation. See these lists for reference:
- UWP: https://docs.microsoft.com/en-us/uwp/api/windows.ui.text.core.coretextinputscope?view=winrt-22000
- Android: https://developer.android.com/reference/android/widget/TextView#attr_android:inputType
I think for a proper implementation of this enum, there's a lot of groundwork that'll need to be done. A better name for this enum may be something more like `Input/TextContext`, and it may be good to limit it to `Text`, `Number`, and maybe `Password` until Android and UWP implementations of this are added.

You also still need to do a pass on naming convention, and have some structs that need moved out of stereokit.h

Faolan-Rad:
I made the enum a bit easier to use

PR #274, opened 2022-01-07
Fix docs folder items shifted by expander arrow

opl-:
The expander arrows were shifting the folder names and annoying me, so here's a fix. Before and after:

comparison

Wasn't sure if this should target master or develop, so I chose develop. I can change it if you want me to.

### DIFF
--- a/docs/src/docs.js
+++ b/docs/src/docs.js
@@ -67,20 +67,32 @@ function renderFolder(folder, parent, hierarchy) {
     folder.element = folderItem;
     parent.appendChild(folderItem);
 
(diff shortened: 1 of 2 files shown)


maluoi:
Hah, this is awesome! Actually, if you want to submit this to master so it goes live right away, that'd be fantastic :)

opl-:
Made a small adjustment to fix keyboard navigation to bring it back to what it was, and added a visual focus indicator for the arrow.

maluoi:
Thank you! That looks so much more clean :)

PR #276, opened 2022-01-10
Upgrade miniaudio to v0.11.3

opl-:
This version of miniaudio fixes a Linux specific bug which (practically speaking) prevents the use of microphone: https://github.com/mackron/miniaudio/issues/376

I had to remove the calls to `ma_decoder_init_memory_raw` in `asset_types/sound.cpp` because that API has changed and no longer seems to support raw as an encoding type. I'm not entirely sure why those calls were there in the first place, as `sound_type_buffer` never seems to make use of the `decoder` anyway. miniaudio provides their own audio buffers, but I opted not to use them since StereoKit reimplements them with the `ring_buffer` type anyway.

I also made the code log errors on miniaudio failures which I think is useful case of failure, but also potentially horribly spammy. Not sure how you'd prefer to handle this, so I'm just submitting this as is.

These changes have been tested on my Debian machine running PulseAudio using the following methods:

- Playing microphone audio in the world (`sound_play(mic_get_stream(), (vec3) {0.0, 0.0, 0.0}, 1.0);`).
- Interacting with the UI to play its interaction sounds which use `sound_generate`.
- Playing an mp3 file: `sound_play(sound_create("/path/audio.mp3"), (vec3) {10.0, 0.0, 0.0}, 1.0)`.


maluoi:
This is also awesome, thank you!

PR #303, opened 2022-02-11
Fix/Release MeshRef

Strepto:
Small fix for an issue causing no meshes to be released.


maluoi:
Thank you! Good catch :)

PR #306, opened 2022-02-13
Call assets_releaseref_threadsafe only on non zero instance pointer

minorai:
Bugfix. It seems like Mesh release reference is only called on IntPtr.Zero, when it should be the other way around.


minorai:
I see this already fixed in #303

maluoi:
Thanks for the PR anyhow! :)

#303 managed to make it into the preview.3 build on NuGet, so the fix is up and available!

PR #348, opened 2022-04-02
Specify requirement for VS2022 and cmake

Strepto:
Just a quick update. I also think I needed to install python 3.x, but cannot remember why. Probably something with OpenXR building.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -13,10 +13,11 @@ The test project is pretty spiffy, and it can be nice to see it in action! There
 StereoKit is a cross-platform library, but many features are not platform specific! If you don't need to test on a platform other than Windows, then you only need some of the dependencies.
 
 **Requisites:**
-- Visual Studio 2019 (2022 may also work)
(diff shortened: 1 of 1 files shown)


maluoi:
Thanks for this! I've been shifting around a bit of the build pipeline this week, VS 2022 is new as of Wednesday, and I'm still looking into switching to vcpkg for some of the Visual Studio build process. So might be more changes there in the near future!

Strepto:
Yeah, I noticed! I actually had a branch for VS 2022 in my fork, but it felt a bit hacky so I did not submit a PR. 
(https://github.com/Strepto/StereoKit/commit/b5cc87a888da069f03539ff4d8f0945c8a29b12f)

maluoi:
Woooh yeah you sure did, I'm glad you were able to navigate through all of that! :D Pretty sure I managed to resolve all those issues with the recent revisions, but feel free to let me know if there's anything else that trips things up :)

Issue #349, opened 2022-04-02
Docs: Consider documenting / explicit mention of the effect of multiplying a bounds with a float


Strepto:
https://github.com/maluoi/StereoKit/blob/dcd95fdfe6bbbd89e2d33ad7c628d97491ad5213/StereoKit/Math/Bounds.cs#L104

This scales the bounds by the given amount, but its not documented with XML or [website](https://stereokit.net/Pages/Reference/Bounds.html). I had to check the source for the definition of what the operator did. 

Usecase: 
I was looking for a scale method to use in the example below, but could not find anything before I found the undocumented multiplier by chance. I tried multiplying with a Vec3 at first, but that did not work. I only kept looking for this instead of fixing my own since most things have a nice intuitive solution in SK ✨ 

```cs
        var scale = 0.1f;
        UI.Handle("1", ref cubePose, Mesh.Cube.Bounds * scale );
        Mesh.Cube.Draw(Material.Default, cubePose.ToMatrix(scale));
```

I would maybe consider a `bounds.WithScale(float scale) -> new scaled bounds` to make it more discoverable that such a method exits, and/or add a xml doc for the existing operator. Also a non-uniform scaling by a Vec3 could maybe be useful, but not needed by me.

Let me know if you want a contribution for this.

maluoi:
Yeah, these are indeed excellent ideas! Adding some methods for this would definitely be a great route for operator discovery :) The current naming convention for this type of thing in StereoKit is `Scale` for self modification, and `Scaled` for returning a modified value (See `Vec3.Normalize` and `Vec3.Normalized`).

Documenting the operator overload would also be really good! One reason I haven't done this yet (for any operators, actually!) is because StereoKit's documenter does not yet support operator overloads, so this is something that needs to be implemented.

If you wanted to write `Scale` and `Scaled` methods with variants+operator for non-uniform, I'd definitely be grateful :) Should probably skip on commentdocs for the operators until SK's documenter supports it properly though, I'll try and implement that soon.

PR #351, opened 2022-04-03
Add Scaled and Scale to Bounds

Strepto:
I tried my way at #349, but not sure if its within the expected style. Also the tests became a bit verbose, but this code should probably not change that much.

Feel free to discard this PR if it adds stuff you do not want to maintain. Also feel free to do whatever adjustments changes you feel like. 👍 

Related to #349

### DIFF
--- a/StereoKit/Math/Bounds.cs
+++ b/StereoKit/Math/Bounds.cs
@@ -1,4 +1,4 @@
-﻿using System.Runtime.CompilerServices;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 namespace StereoKit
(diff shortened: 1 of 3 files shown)


StereoKit/Math/Bounds.cs
```diff
@@ -101,7 +101,41 @@ public bool Contains(Vec3 linePt1, Vec3 linePt2)
...
+			center *= scale;
+			dimensions *= scale;
+		}
+
+		/// <summary>Scale the bounds position and extents by the given scale.
+		/// Returns a new Bounds.</summary>
+		/// <remarks>This is equivalent to using the * operator: bounds * scale</remarks>
+		/// <param name="scale">Scale</param>
+		/// <returns>A new scaled bounds</returns>
+		[Pure]
```
Strepto, line 127: I'm not sure on my use of the [Pure] here as the attribute is never validated by the compiler, and not a guarantee. Feel free to remove it if its not fitting the style.

maluoi: Reading up on Pure, it sounds like they may be a little archaic, but [will produce warnings](https://www.infoq.com/news/2019/01/pure-attribute-net-core/) with some code analyzers if the return value of the Method is ignored! If so, that sounds like a great reason to use it :)

maluoi:
This looks perfect to me, thank you so much!

Not a dealbreaker here, I'm not a big fan of `var` for readability. If it was in public docs or the StereoKit project, I'd probably want it changed, but it's totally fine in the tests :)

PR #353, opened 2022-04-04
Fix typos & grammar in doc/UI/Input

nschapeler:


maluoi:
Thanks so much for this! FYI, StereoKit's documentation is auto-generated from the commentdocs during version releases, so you'd want to [make your modifications over here](https://github.com/maluoi/StereoKit/blob/develop/StereoKit/Systems/UI.cs#L432). Changes should also be applied to the `/develop` branch! I generally treat `/master` as a snapshot of the latest full release.

It looks like I may have at least updated the spelling on that one since v0.3.5, but it is indeed still a pretty rough sentence!

nschapeler:
Was unaware that's how you had the workflow setup, thanks for letting me know! :) Since you already fixed it in develop, I'll close this PR.

Issue #356, opened 2022-04-06
Providing Eye Gaze Sample Time


sandrist:
The [OpenXR spec](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_EXT_eye_gaze_interaction) indicates that it is possible to query the actual sample time for a given eye gaze pose, which is useful for reasoning about high accuracy eye tracking, computing precise velocities over time, etc. It would be great if StereoKit could also provide that information.

I've implemented and locally tested one possible approach, which I was wondering if you might be willing to provide feedback on in a PR? I'm open to alternative approaches to surfacing this information, of course.

Basically, we just need a slightly modified version of `openxr_get_space()` to get the gaze space along with this extra timing information (it's specific to just gaze, and doesn't seem to be anything analogous for the head or hands).

```csharp
bool32_t openxr_get_gaze_space(pose_t* out_pose, XrTime &out_sample_time, XrTime time) {
	if (time == 0) time = xr_time;

	XrEyeGazeSampleTimeEXT gaze_sample_time = { XR_TYPE_EYE_GAZE_SAMPLE_TIME_EXT };
	XrSpaceLocation space_location = { XR_TYPE_SPACE_LOCATION, &gaze_sample_time };
	XrResult        res = xrLocateSpace(xr_gaze_space, xr_app_space, time, &space_location);
	if (XR_UNQUALIFIED_SUCCESS(res) && openxr_loc_valid(space_location)) {
		memcpy(&out_pose->position, &space_location.pose.position, sizeof(vec3));
		memcpy(&out_pose->orientation, &space_location.pose.orientation, sizeof(quat));
		out_sample_time = gaze_sample_time.time;
		return true;
	}
	return false;
}
```

Then this sample time can be accessible in `Input.cs` as something like `Input.EyesSampleTime`.

Thoughts?

sandrist:
One immediate question is how to fill in that information when eye gaze is simulated in a "flatscreen" mode... Perhaps it should indicate the same as OpenXR.Time? Or just a default value (0)?

maluoi:
Okay, this is certainly interesting! The general approach you've outlined here seems reasonable, but gets a bit sticky when referring to time.

[...]

PR #358, opened 2022-04-08
Added `Backend.OpenXR.EyesSampleTime`

sandrist:
Addresses Issue #356

[OpenXR](https://www.khronos.org/registry/OpenXR/specs/1.0/html/xrspec.html#XR_EXT_eye_gaze_interaction) provides the ability to query the actual device sample time associated with a given gaze tracker result. This PR surfaces that information so it is accessible in StereoKit.

- Created a separate `openxr_get_gaze_space()` method that is similar to the existing `openxr_get_space()` but fills in the extra timing information specific to gaze.
- Added `Backend.OpenXR.EyesSampleTime` for querying that sample time.
- Added an example of how to use this information in the existing `DemoEyes.cs`. The sample time is used to determine and display the true sampling frequency of the gaze tracker.


maluoi:
GitHub is reporting a conflict here. But other than that, everything looks good!

maluoi:
Okay, nevermind, that was just the rebase option!

PR #359, opened 2022-04-13
Fix GLXDrawable in graphics binding

ChristophHaag:
skg_get_platform_data stores a pointer to glXDrawable

    result._glx_drawable  = &glxDrawable;

therefore we need to dereference the pointer when putting the
GLXDrawableinto value into the graphics binding


maluoi:
Thanks for tracking this down! Another one of those bugs where I wonder how things have been working for so long with this present.

Could you target this towards `/develop` instead? For StereoKit, `/master` is intended to be a fixed snapshot of the latest full version.

maluoi:
Oh sweet, I can change the merge branch from my end! No worries then, thanks again!

technobaboo:
hahahaha yea how did that work? i'm surprised I got the graphics to work at all tbh, no idea how this worked lol

PR #370, opened 2022-05-10
fix(C API): misplaced endif from line 418 to line 441

MalekiRe:


maluoi:
Should already be fixed in `/develop`, thank you! :)

PR #378, opened 2022-05-31
Install StereoKitCTest by default in CMake

slitcch:


maluoi:
Not certain here, should this be gated by `SK_BUILD_TESTS`, or does that get handled properly automatically when the `StereoKitCTest` target doesn't exist?

slitcch:
Oh hmm, let me try that.

maluoi:
I imagine there may also be some work related to the Assets the app uses, the cmake file doesn't actually mention those anywhere.

slitcch:
Yep, good catch :)

slitcch:
Not high enough priority for me to bikeshed people on this, it's fine the way it was

PR #420, opened 2022-07-03
Tell flatscreen_input that we can't move the pointer under XWayland

slitcch:


maluoi:
Looks pretty good to me, thanks for this! :)

PR #428, opened 2022-07-08
- Make pointer warping work under XWayland

slitcch:


maluoi:
Okay, that looks pretty bizarre! But if it works, it works! Thanks for digging through this :)

PR #443, opened 2022-07-29
Making HolographicTheme compile on .NET Standard 2.0

Zee2:
`MathF` is unavailable on .NET Standard 2.0, which is required when targeting UWP, so this changes `MathF` calls to StereoKit `SKMath` calls! Ez.


maluoi:
Ty! :)

PR #454, opened 2022-08-12
Header define for FFI that can remove unnecessary functions

technobaboo:
I noticed rust had a ton of warnings due to functions being included in the header that were very unnecessary for ffi so I just added an #ifdef for SK_FFI to not include some headers and define char types


slitcch:
LGTM

PR #458, opened 2022-08-18
Fix for EGL crashing in XR mode on Linux

technobaboo:
just don't start the x window in XR mode, EGL is useful because you don't need to make a window so standalone headsets can run XR apps

### DIFF
--- a/StereoKit/Native/NativeTypes.cs
+++ b/StereoKit/Native/NativeTypes.cs
@@ -66,6 +66,12 @@ namespace StereoKit
 		/// You don't want this, you can disable it with this setting!</summary>
 		public  bool disableFlatscreenMRSim { get { return _disableFlatscreenMRSim > 0; } set { _disableFlatscreenMRSim = value ? 1 : 0; } }
 		private int _disableFlatscreenMRSim;
+		/// <summary>By default, StereoKit will open a desktop window for
+		/// keyboard input due to lack of XR-native keyboard APIs on many
+		/// platforms. If you don't want this, you can disable it with
+		/// this setting!</summary>
+		public  bool disableDesktopInputWindow { get { return _disableFlatscreenMRSim > 0; } set { _disableFlatscreenMRSim = value ? 1 : 0; } }
+		private int _disableDesktopInputWindow;
 		/// <summary>By default, StereoKit will slow down when the
 		/// application is out of focus. This is useful for saving processing
 		/// power while the app is out-of-focus, but may not always be
--- a/StereoKit/Native/NativeTypes.cs
+++ b/StereoKit/Native/NativeTypes.cs
@@ -70,7 +70,7 @@ namespace StereoKit
 		/// keyboard input due to lack of XR-native keyboard APIs on many
 		/// platforms. If you don't want this, you can disable it with
 		/// this setting!</summary>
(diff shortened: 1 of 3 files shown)


StereoKitC/platforms/linux.cpp
```diff
@@ -518,9 +509,10 @@ void linux_shutdown() {
 ///////////////////////////////////////////
 
 void linux_step_begin_xr() {
-  #if !defined(SKG_LINUX_EGL)
-	linux_events();
+  #if defined(SKG_LINUX_EGL)
```
slitcch, line 512: Not my project and not my style guide but I really feel like you should expand this to `#if` cases and use brackets

StereoKitC/platforms/linux.cpp
```diff
@@ -436,6 +424,11 @@ bool linux_start_pre_xr() {
 ///////////////////////////////////////////
 
 bool linux_start_post_xr() {
+	#if defined(SKG_LINUX_EGL)
+	if (!sk_settings.disable_xr_window && !setup_x_window())
```
slitcch, line 428: Again I wish for brackets

slitcch:
Tested all four codepaths on Monado, all work as expected except for GLX+disable_xr_window=true - in that case I feel like input should be disabled, maybe?

Either way this is good enough and I'm fine with it going in as-is :)

technobaboo:
> 

we can't close the window on GLX or it'd disconnect from the XR runtime though

StereoKitC/platforms/linux.cpp
technobaboo: @maluoi thoughts?

slitcch:
> we can't close the window on GLX or it'd disconnect from the XR runtime though

Yes, but you can just drop input data from linux_events on the floor

technobaboo:
> Yes, but you can just drop input data from linux_events on the floor
why would we want to do that? it makes no sense

StereoKit/Native/NativeTypes.cs
```diff
@@ -66,6 +66,12 @@ public struct SKSettings
 		/// You don't want this, you can disable it with this setting!</summary>
 		public  bool disableFlatscreenMRSim { get { return _disableFlatscreenMRSim > 0; } set { _disableFlatscreenMRSim = value ? 1 : 0; } }
 		private int _disableFlatscreenMRSim;
+		/// <summary>By default, StereoKit will open a desktop window for
+		/// keyboard input due to lack of XR-native keyboard APIs on many
+		/// platforms. If you don't want this, you can disable it with
+		/// this setting!</summary>
+		public  bool disableDesktopInputWindow { get { return _disableFlatscreenMRSim > 0; } set { _disableFlatscreenMRSim = value ? 1 : 0; } }
```
maluoi, line 73: This references `_disableFlatscreenMRSim` instead of `_disableDesktopInputWindow`!

StereoKitC/platforms/linux.cpp
maluoi: This looks kinda gnarly, I'm inclined to agree that this needs re-worked.

slitcch:
> why would we want to do that? it makes no sense
If you set `disable_desktop_input_window` to `true`, it might be because you don't want input from the desktop!

technobaboo:
but if you do want it whenever possible, there's no option for that! it's just either all input or no input when the reason to disable it for EGL is specifically to get rid of the window that doesn't work with... any XR compositor i'll be making

PR #459, opened 2022-08-19
Add tex_t owned struct field to avoid destroying non-stereokit textures

technobaboo:


StereoKitC/asset_types/texture_.h
```diff
@@ -6,7 +6,7 @@ namespace sk {
 
 void        tex_set_zbuffer      (tex_t texture, tex_t depth_texture);
 void        tex_set_options      (tex_t texture, tex_sample_ sample = tex_sample_linear, tex_address_ address_mode = tex_address_wrap, int32_t anisotropy_level = 4);
-void        tex_set_surface      (tex_t texture, void *native_surface, tex_type_ type, int64_t native_fmt, int32_t width, int32_t height, int32_t surface_count);
+void        tex_set_surface      (tex_t texture, void *native_surface, bool32_t owned, tex_type_ type, int64_t native_fmt, int32_t width, int32_t height, int32_t surface_count);
```
maluoi, line 9: This function prototype is actually a duplicate of the one in stereokit.h! This should probably be removed entirely :)

StereoKitC/stereokit.h
```diff
@@ -841,7 +841,7 @@ SK_API tex_t        tex_create_cubemap_files(const char **cube_face_file_xxyyzz,
 SK_API void         tex_set_id              (tex_t texture, const char *id);
 SK_API const char*  tex_get_id              (const tex_t texture);
 SK_API void         tex_set_fallback        (tex_t texture, tex_t fallback);
-SK_API void         tex_set_surface         (tex_t texture, void *native_surface, tex_type_ type, int64_t native_fmt, int32_t width, int32_t height, int32_t surface_count);
+SK_API void         tex_set_surface         (tex_t texture, void *native_surface, bool32_t owned, tex_type_ type, int64_t native_fmt, int32_t width, int32_t height, int32_t surface_count);
```
maluoi, line 844: For public API changes like this, `bool32_t owned` should be the last item in the signature, and should have an `sk_default(true)` that matches the original behavior.

maluoi: This should mean that most of the rest of the changes can be dropped

maluoi (changes requested):

StereoKitC/asset_types/texture_.h
technobaboo: i figured, why is that header file there and can i just yeet the whole header?

StereoKitC/stereokit.h
technobaboo: do you still want to expose that to C#?

StereoKitC/asset_types/texture_.h
maluoi: The header file itself is important, there's some items separated out here into different groups to help with improving build time :)

StereoKitC/stereokit.h
maluoi: Yeah! C# similarly should have a default value for owned, and I generally prefer for the signatures to be in the same order as the C function.

StereoKit/Native/NativeAPI.cs
```diff
@@ -146,7 +146,7 @@ internal static class NativeAPI
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   tex_set_colors          (IntPtr texture, int width, int height, [In] byte[] data);
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   tex_set_colors          (IntPtr texture, int width, int height, [In] ushort[] data);
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   tex_set_colors          (IntPtr texture, int width, int height, [In] float[] data);
-		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   tex_set_surface         (IntPtr texture, IntPtr native_surface, TexType type, long native_fmt, int width, int height, int surface_count);
+		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   tex_set_surface         (IntPtr texture, IntPtr native_surface, TexType type, long native_fmt, int width, int height, int surface_count, bool owned);
```
maluoi, line 149: Oh! This should be `int` for bool32_t

maluoi (changes requested):

Issue #471, opened 2022-09-06
When calling the getter Material on TextStyle, it doesn't increment the ref counter when creating the Material object


smasonuk:
## Description
In this line in Text.cs, it returns a new Materia object, but the ref counter isn't incremented. 
`public Material Material => new Material(NativeAPI.text_style_get_material(this));`

However in the finalizer for Material it decrements the counter:
`NativeAPI.assets_releaseref_threadsafe(_inst);`

## Platform / Environment
Any that use the c# StereoKit library.

PR #472, opened 2022-09-06
Solution for issue #471. Ensure that the ref counter is incremented.

smasonuk:
I created this ticket for this issue: https://github.com/StereoKit/StereoKit/issues/471

And further details also in discussion https://github.com/StereoKit/StereoKit/discussions/463

This is my first PR, not sure if I'm doing this right!

### DIFF
--- a/StereoKit/Native/NativeAPI.cs
+++ b/StereoKit/Native/NativeAPI.cs
@@ -188,6 +188,7 @@ namespace StereoKit
 
 		///////////////////////////////////////////
 
+		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   material_addref          (IntPtr material);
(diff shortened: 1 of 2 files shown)


maluoi:
Thanks for looking into this, and sorry for the slow reply, I'm technically on vacation for just a _little_ bit longer 🙃 

It definitely looks like you've hit upon the root of the issue here! However, I would consider the reference counting issue you've spotted to be at the C level rather than the C# level, so the fix should be over here: https://github.com/StereoKit/StereoKit/blob/develop/StereoKitC/systems/text.cpp#L171 ! Otherwise, C/C++ developers will also encounter approximately the same issue :)

smasonuk:
No worries, I did consider doing it there but not as confident with C :-)

PR #476, opened 2022-09-18
Direction Vector Math

spatialfree:
A simple and useful piece of vector math that is easy to get backwards without this chunk of code ^-^


maluoi:
Oh I like this, excellent idea, thank you!

PR #479, opened 2022-10-02
Relative and Delta Quaternion Math

spatialfree:
Quaternion methods and conventions I've found invaluable and would like to showcase as StereoKit functionality in my upcoming FOSSXR talk!

I would also like to Deprecate: Quat.Difference(a, b)
as it doesn't clearly communicate the "direction" of the resulting delta through its operator names and obfuscates the underlying math of (delta = to - from)

### DIFF
--- a/Examples/StereoKitTest/Docs/DocQuat.cs
+++ b/Examples/StereoKitTest/Docs/DocQuat.cs
@@ -2,7 +2,21 @@
 
 class DocQuat : ITest
 {
-	public void Initialize() { }
+	Pose spherePose;
+	Quat sphereDelta;
+	Vec3 oldPalmPos;
+
+	Material matDev;
+
+	public void Initialize() 
+	{
+		spherePose = new Pose(-1,0,1, Quat.Identity);
+		sphereDelta = Quat.Identity;
+
+		matDev = Material.Default.Copy();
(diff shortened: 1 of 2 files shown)


maluoi:
This looks pretty fantastic to me!

One change I'd like to see here though, is that StereoKit already lists arguments in "from, to" order, (see: [`Quat.LookAt`](https://stereokit.net/Pages/StereoKit/Quat/LookAt.html)) rather than "to, from". If you could swap that around, I believe that'd make things more consistent :)

I would _love_ to see sample snippets for this as well, (like [over here](https://github.com/StereoKit/StereoKit/blob/master/Examples/StereoKitTest/Docs/DocQuat.cs) or something) but, no worries if you're not into that quite yet!

spatialfree:
Been doing all this without building the project so far 😅 as coming from Unity it's a lot to take in. Though I did my best to test the code sample in my own project before plugging it in!

maluoi:
I made some small changes here to get it to compile! Something related to operator confusion when converting to and from System.Numerics quaternions, but no biggie :)

I also reorganized a little of the sample code so variable definitions could fit in with the sample code too. Also, that sample is _fantastically_ cool, thank you so much for that!

spatialfree:
Thanks! I also looked through your changes to understand how to better contribute next time 🍉

PR #495, opened 2022-10-18
Multimaterial skinned meshes

cameronfr:
In multimaterial skinned meshes, the GLTF skin node has multiple primitives.

This solution feels kinda hack-y, maybe the `node_map` should be replaced with a `hashmap_t<cgltf_node*, hashmap_t<cgltf_size, model_node_id>>`, since GLTF nodes with multiple primitives correspond to multiple `sk_mesh` nodes? (Not sure how to free that nicely, though). 

If a GLTF skin node has a mesh primitive that fails to parse, this code will fail since the offset will be wrong (although I think the code would fail before too, if a skin's mesh failed to parse)

I can refactor or this can just serve as reference!


maluoi:
Hey thanks so much for this! I'm happy this wasn't too much code, and I like what I see so far! I'll test this out in-depth a little later this week, and I'll definitely make sure this is included with the next build :) I imagine it won't take long for me to review, I'm just in deep with some MSBuild stuff right this moment.

PR #513, opened 2022-11-08
Feature/backend openxr pollevent

jackdaus:
Implemented the Backend.OpenXR.PollEvent callback for #493! Let me know if you want me to make any changes or add any specific test(s). 

It was pretty tricky figuring out the C#/C interop since I didn't have much experience with that, but I think I got it figured it out. I did some local testing, and adding/removing callbacks seems to work as expected. 

I also wanted to make sure this PR would work with my larger goal of implementing the FBSpatialEntity extensions, so I got a little initial demo of that over here: https://github.com/jackdaus/StereoKit/tree/feature/fb_spatial_entity_ext. It's just got the initial create anchor extension working, but I am planning on implementing most of the other pieces of this extension soon (like storing and querying anchors). I could also make a PR with that branch eventually if you think it's worth adding to SK.

### DIFF
--- a/StereoKitC/xr_backends/openxr.cpp
+++ b/StereoKitC/xr_backends/openxr.cpp
@@ -45,6 +45,11 @@ typedef struct context_callback_t {
 	void *context;
 } context_callback_t;
 
+typedef struct poll_event_callback_t {
+	void (*callback)(void* XrEventDataBuffer);
+	void *context;
+} poll_event_callback_t;
+
 XrFormFactor xr_config_form = XR_FORM_FACTOR_HEAD_MOUNTED_DISPLAY;
 const char *xr_request_layers[] = {
 	"",
@@ -71,7 +76,8 @@ array_t<const char*> xr_exts_user    = {};
 array_t<uint64_t>    xr_exts_loaded  = {};
 bool32_t             xr_minimum_exts = false;
 
-array_t<context_callback_t> xr_callbacks_pre_session_create = {};
+array_t<context_callback_t>    xr_callbacks_pre_session_create = {};
+array_t<poll_event_callback_t> xr_callbacks_poll_event         = {};
 
 bool   xr_has_bounds        = false;
 vec2   xr_bounds_size       = {};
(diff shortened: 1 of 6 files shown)


maluoi:
Fantastic, this is looking really great! I'll have a few notes for you over on the code in a moment, and I'll try and give this a spin tomorrow :)

StereoKitC/xr_backends/openxr.cpp
```diff
@@ -697,6 +703,11 @@ void openxr_poll_events() {
 			break;
 		default: break;
 		}
+
```
maluoi, line 706: I'm curious if you considered putting this 'before' SK acts on the event? (my gut says 'after' is more correct, but I haven't gotten too deep here)

Also wonder what your thoughts are on having a system for consuming events, and not allowing them to fall through? I know some window event loops have this mechanism to say "I handled it already, you don't need to!", but I don't know if that'll actually be necessary here.

jackdaus: Hmm so I haven't given this a _whole_ lot of thought, but one thing I considered is that allowing the user's callback to act before SK could cause this issue: since I am passing a reference of `event_buffer` to the user's callback via `&event_buff`, the user could modify (maybe inadvertently) the `event_buffer` struct data and cause issues when it's SK's turn to process the event. So that was my reasoning there - to avoid the user messing up the `event_buffer` data before SK got a chance to process it.

As for consuming an event and not allowing them to fall through, I guess it could be a useful feature for some folks? But I can't think of a particular scenario off the top of my head - which is probably just because I don't have much experience with OpenXR yet. But there could be good use cases! I just can't think of any.

From my perspective, I think I'd probably just be using the add callback feature to add event handlers for events that _aren't_ already handled by SK (like events coming from OpenXR extensions).  So the option to prevent fall through is a moot point to me. But I could definitely add the option if you'd like! Let lmk

maluoi: Okay, I think I'm satisfied with this for the moment! Leave it as it is, and we can make changes later if we find it necessary :)

jackdaus: Okey doke!

jackdaus:
Added `context` to the callbacks! Just let me know if you want any other changes

maluoi:
Fantastic work once again! Thanks for this one, seems to work perfectly for me!

Issue #538, opened 2022-12-01
Thread-safe Stepper add/remove


maluoi:
Adding and removing `IStepper`s from a separate thread can cause issues. Here's a small repro case where `Step` will sometimes get called after `Shutdown`. Some initial discussion and related work happened in #528.

```csharp
using StereoKit;
using StereoKit.Framework;
using System;
using System.Collections.Generic;
using System.Threading.Tasks;

SK.Initialize("Stepper Stress Test");

Task.Run(() => {
	Random r = new Random();
	List<StepperTest> steppers = new List<StepperTest>();
	while (true)
	{
		for (int i = 0; i < 10000; i++)
			steppers.Add(SK.AddStepper<StepperTest>());

		for (int i = 0; i < 10000; i++)
			steppers.RemoveAt(r.Next(steppers.Count));
	}
});

SK.Run(() => {
	Mesh.Sphere.Draw(Material.Default, Matrix.TS(0, 0, -0.5f, 0.1f));
});

class StepperTest : IStepper
{
	private Model model;
	public  bool  Enabled => true;

	public bool Initialize()
	{
		Model m = new Model();
		return true;
	}
	public void Shutdown() => model = null;
	public void Step    () => model.Draw(Matrix.Identity);
}
```

One proposal was to shift the stepper list over to a `ConcurrentDictionary` (maybe `ConcurrentBag`?). Another possibility could be some separate structure that's added to from parallel threads, then executed at a specific moment on the main thread.

austinbhale:
I started some work on this issue here https://github.com/StereoKit/StereoKit/compare/develop...austinbhale:StereoKit:develop and found that using a `BlockingCollection` to protect access to the list kept most of the expected Steppers functionality when being added/removed/stepped through.

---
The other Concurrent classes were very tricky to implement in this scenario:

### ConcurrentStack / ConcurrentBag / ConcurrentQueue
Removing a specific element from a bag or stack requires you to take items out of the collection and put them back in again. The `GetEnumerable` method creates a snapshot in time, so anything modified in the collection is not seen (https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1.getenumerator?redirectedfrom=MSDN&view=net-7.0#remarks). If you were to iterate over this collection in `Step`, and a `Shutdown` was called for an item, the stepper will still call its own `Step` after a `Shutdown` call.

### ConcurrentDictionary
[...]

PR #542, opened 2022-12-05
Thread safe steppers

maluoi:
Discussion and details over in #538! Thank you @austinbhale for working through this with me, if you spot anything else that should be changed, let me know!

### DIFF
--- a/StereoKit/Framework/Steppers.cs
+++ b/StereoKit/Framework/Steppers.cs
@@ -6,9 +6,16 @@ namespace StereoKit.Framework
 {
 	internal class Steppers
 	{
-		List           <IStepper> _steppers = new List<IStepper>();
-		ConcurrentQueue<IStepper> _remove   = new ConcurrentQueue<IStepper>();
-		ConcurrentQueue<IStepper> _add      = new ConcurrentQueue<IStepper>();
+		enum   ActionType { Add, Remove }
+		struct StepperAction
+		{
+			public IStepper   stepper;
+			public ActionType type;
+			public StepperAction(IStepper stepper, ActionType type) { this.stepper = stepper; this.type = type; }
+		}
+
+		List           <IStepper>      _steppers = new List<IStepper>();
+		ConcurrentQueue<StepperAction> _actions  = new ConcurrentQueue<StepperAction>();
 
 		public void Shutdown()
 		{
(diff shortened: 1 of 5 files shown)


StereoKit/Framework/Steppers.cs
```diff
@@ -1,74 +1,85 @@
...
 			{
-				_steppers[i].Step();
+				switch (action.type)
+				{
+					case ActionType.Add:
+						action.stepper.Initialize();
+						_steppers.Add(action.stepper);
+						break;
+					case ActionType.Remove:
+						_steppers.Remove(action.stepper);
```
austinbhale, line 68: How about shutting down only if it is found in the list?
(e.g., two threads try to remove a generic type's steppers at ~nearly the same time, so Shutdown isn't called twice for the same stepper)

austinbhale (changes requested):
That's some nice looking code right there! I've attached one small comment while thinking of potential edge cases.

StereoKit/Framework/Steppers.cs
austinbhale: `if (_steppers.Remove(action.stepper)) action.stepper.Shutdown();`

StereoKit/Framework/Steppers.cs
```diff
@@ -1,74 +1,85 @@
...
+				}
 			}
+
+			foreach (IStepper stepper in _steppers)
+				stepper.Step();
+		}
+
+		public void Shutdown()
+		{
+			_actions = new ConcurrentQueue<StepperAction>();
```
austinbhale, line 80: If shutdown is called early, we should give some indicator to halt the `Step` or we get the following exception:
image
I'll come up with an idea and push to this branch, if that's okay

maluoi: Oh interesting! I think I'd consider this to be an invalid use of `SK.Shutdown`, and I may want to add some guards to prevent it from happening? I'll probably throw an exception if called while SK is still running. `SK.Quit` is the correct function to call in this case.

StereoKit/Framework/Steppers.cs
maluoi: Good catch, yeah, probably similar if a stepper's initialization fails.

StereoKit/Framework/Steppers.cs
austinbhale: `SK.Quit` works perfectly with this test case. Cool, so `Quit` must be called to step through the last frame, then it shutdowns. An exception would be great!

maluoi: Exception added :)

maluoi:
Awesome! Thanks so much for your work and reviews on this one as well! Good stuff :)

Issue #526, opened 2022-11-14
Texture FromFile is blocking on the main UI thread


laultman:
## Description
I am loading a Dictionary KVP with the textures from a collection of .png files. This runs as a Task.Run. Inside the task the code paginates the texture into block of files with the idea the app could access textures as soon as the first block loads.
The code calls the method stemming from the SK Initialize method. Since Initialize is not async the method calls a method to launch my async code by calling an async void Player method. The Player does some SK initializations and calls an async Task loader to get the textures.
The app is using remoting to the HL2. The app starts and the Receiving ... shows in the HL2 It disappears after a few seconds and all is normal. In the debugger all the async code runs to completion as expected.
My main menu has a button for a player. When this button is pressed to enable a block of UI instructions. The debugger shows the enable flag is not set which is all the button does. What happens is the HL2 goes blank (the remote stops streaming) for about 10-15 seconds and then resumes streaming. I am at a loss here.
The code that is called by the button push is in its own Step having been derived from IStepper.

## Platform / Environment
Windows 11 Surface Pro 7 intel i7 16g ram.

## Logs or exception details
```
no errors, no messages on the diagnostics window.
```
_If you have an exception, details about that would also be essential._

maluoi:
Texture loading is already async, so loading them async is probably not terribly helpful here! In addition to this, accessing certain parameters before they're loaded can cause the application to pause until that content is loaded and accessible.

So like:
```csharp
SK.Initialize();

string[]   files     = new string[] { "1.png", "2.png", "3.png" };
Material[] materials = files.Select( f => {
    Material result = Material.Default.Copy();
    result[MatParamName.DiffuseTex] = Tex.FromFile(f);
    return result;
});

// DON'T do these here, they will hang the app until colors/width are available
// material[0][MatParamName.DiffuseTex].GetColors()
// material[0][MatParamName.DiffuseTex].Width

int curr = 0;
SK.Run(() => {
    int idx = curr % materials.Length;
    curr++;

    // This will draw with "fallback" textures before they've loaded
    Mesh.Cube.Draw(materials[idx], Matrix.T(-1,0,0));
    
    // This will only draw when all assets are loaded
    if (Assets.CurrentTask >= Assets.TotalTasks) {
        Mesh.Cube.Draw(materials[idx], Matrix.T(0,0,0));
    }
    
    // This will only draw materials where the texture is fully loaded
[...]

PR #574, opened 2023-01-16
Adds a synchronous memory load for textures.

austinbhale:
Hi nick, I have a stream of JPEG encoded images coming from another thread and found the asynchronous loading system difficult to get _just right_. I tried using the `OnLoaded` events to synchronize to the SK game thread as well as attempts at your suggestions here about checking the `AssetState`: https://github.com/StereoKit/StereoKit/issues/526#issuecomment-1314510572. At the end of the day, it was much simpler to have a `Tex.SetMemory` that performs synchronously, like `Tex.SetColors`. Here's it in practice with a JPEG-encoded MediaCapture stream on the HL2: 

https://user-images.githubusercontent.com/32556981/212746350-51c88027-bbd9-4473-8fcc-3c01077335dc.mp4

I created a new method for `tex_set_arr_parse` to separate itself from any asset loading. Curious about your thoughts on this! Would be happy to change anything, I'm just glad it's working now :)

### DIFF
--- a/StereoKit/Assets/Tex.cs
+++ b/StereoKit/Assets/Tex.cs
@@ -263,25 +263,25 @@ namespace StereoKit
 			NativeAPI.tex_set_colors(_inst, width, height, data);
 		}
 
-        /// <summary>Loads an image file stored in memory directly into
-        /// the created texture! Supported formats are: jpg, png, tga,
-        /// bmp, psd, gif, hdr, pic. This method introduces a blocking 
-        /// boolean parameter, which allows you to specify whether this
-        /// method blocks until the image fully loads! The default case
-        /// is to have it as part of the asynchronous asset pipeline, in
-        /// which the Asset Id will be the same as the filename.</summary>
-        /// <param name="imageFileData">The binary data of an image file,
-        /// this is NOT a raw RGB color array!</param>
-        /// <param name="sRGBData">Is this image color data in sRGB format,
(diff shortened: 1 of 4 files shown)


maluoi:
I think this should have an option to be synchronous or not, and should by default be async! And I would probably like to refer to it as `blocking` just to make it explicit. The big time cost that _really_ makes async necessary is the format decode, and jpeg in particular can be sloooow.

I also have future plans to add texture compression as part of this pipeline, so I want to be thinking about that here as well. A worst case call here could be synchronous decode of a large jpeg, calculation of mip-maps, and conversion to a compressed format. I wouldn't want people to stumble into that by accident 😬

austinbhale:
> I think this should have an option to be synchronous or not, and should by default be async! And I would probably like to refer to it as `blocking` just to make it explicit. The big time cost that _really_ makes async necessary is the format decode, and jpeg in particular can be sloooow.

Great suggestion, I changed it to an optional parameter and reintroduced the priority for the async asset loading system

> I also have future plans to add texture compression as part of this pipeline, so I want to be thinking about that here as well. A worst case call here could be synchronous decode of a large jpeg, calculation of mip-maps, and conversion to a compressed format. I wouldn't want people to stumble into that by accident 😬

Yikes! There's a major bottleneck for the default case, yea I think the default should be async for most use cases. Though having it blocking made my life a lot easier from a multithreaded pov :)

Is there anything else you'd like to have changed?

PR #577, opened 2023-01-19
feat: implement an `Matrix.Perspective` overload for camera perspectives

austinbhale:
This uses the same calculation as `CreatePerspectiveFieldOfView` but takes advantage of the camera intrinsic information for its near plane dimensions (width + height).
I can prove that this works, if needed, but also here's a similar calculation by MSFT: https://learn.microsoft.com/en-us/previous-versions/windows/desktop/bb281732(v=vs.85)

### DIFF
--- a/StereoKit/Math/Matrix.cs
+++ b/StereoKit/Math/Matrix.cs
@@ -417,12 +417,12 @@ namespace StereoKit
 		/// <returns>The final perspective matrix.</returns>
 		public static Matrix Perspective(Vec2 imageResolution, Vec2 focalLength, float nearClip, float farClip)
 		{
-		    Vec2 aspectRatio = imageResolution / focalLength;
+		    Vec2 nearPlaneDimensions = imageResolution / focalLength;
 		    float deltaZ = nearClip - farClip;
 		    return new Matrix4x4()
 		    {
-			M11 = 2 / aspectRatio.x,
-			M22 = 2 / aspectRatio.y,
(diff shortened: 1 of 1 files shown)


maluoi:
This looks somewhat similar to [`Matrix4x4.CreatePerspective`](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.matrix4x4.createperspective?view=net-7.0) as well, I think? Does `imageResolution` and `focalLength` make this totally incompatible, or can those calculations be fed in to `Matrix4x4.CreatePerspective`?

austinbhale:
NearPlaneDimensions var should be the width/height on the unit plane, so I think that might work! In transit atm, but I’ll test it out shortly

austinbhale:
Yep nice find! This produces the same results:
```
Vec2 nearPlaneDimensions = imageResolution / focalLength * nearClip;
return Matrix4x4.CreatePerspective(nearPlaneDimensions.x, nearPlaneDimensions.y, nearClip, farClip);
```

So would you want to overload the `Matrix4x4.CreatePerspective` method in SK (with or w/o the focal length & image resolution), or should I keep this in my own code?

maluoi:
I'd be pretty happy to have it as a small override for `Matrix.Perspective`!

austinbhale:
Sorry for the delay; how does the latest update look?

maluoi:
Looks good to me!

PR #578, opened 2023-01-20
Added model_recalculate_bounds_exact() 

paulmelis:
This computes a model bound based on (transformed) vertices, instead of using per-mesh AABBs. Main reason for this is to have a much tighter bound, especially for interaction volumes, at the cost of a longer time needed to compute bounds for models with many vertices.

Note that I haven't tested this from the C# side, but the C code does compile on Linux :smile:

### DIFF
--- a/StereoKit/Assets/Model.cs
+++ b/StereoKit/Assets/Model.cs
@@ -419,6 +419,13 @@ namespace StereoKit
 		public void RecalculateBounds()
 			=> NativeAPI.model_recalculate_bounds(_inst);
 
+		/// <summary>Examines the visuals as they currently are, and rebuilds
+		/// the bounds based on all the vertices in the model! This leads (in general)
+		/// to a tighter bound than the default bound based on bounding boxes.
+		/// However, computing the exact bound can take much longer!</summary>
+		public void RecalculateBoundsExact()
+			=> NativeAPI.model_recalculate_bounds_exact(_inst);
+
 		/// <inheritdoc cref="Model.Draw(Matrix)"/>
 		/// <param name="colorLinear">A per-instance linear space color value
 		/// to pass into the shader! Normally this gets used like a material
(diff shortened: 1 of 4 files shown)


paulmelis:
Btw, `model_recalculate_bounds_exact()` isn't called in any way, except when forced through `Model.RecalculateBoundsExact()`. Perhaps that call can even be offloaded to a thread/asyn, in order not to lock up the app.

paulmelis:
Comparison on a model file I noticed the (still) quite loose bounds:

Current: center [-0, 0.26, 0], dimensions [12.43, 0.52, 11.94]

image

Exact: center [-4.64, -0, -4.25], dimensions [4.64, 0.52, 4.25]

image

maluoi:
This is a pretty neat feature! Do you know what level of "slower" this might be? I imagine it's probably not so bad that it'd really need to be async, but I'm curious!

I've also taken to using some of the [DirectXMath functions directly for bounds stuff](https://github.com/StereoKit/StereoKit/blob/develop/StereoKitC/intersect.cpp#L147), I think its likely to have some good speed benefits in this loop! `XMVectorMin/Max` in particular, in case the compiler isn't auto-vectorizing that already. It'd be nice if you could add that here! Might also be worth hoisting the `XMMATRIX` conversion out to a higher scope [like here](https://github.com/StereoKit/StereoKit/blob/1bf86b281e883041f9e0dbba32259de47966b46d/StereoKitC/systems/sprite_drawer.cpp#L79), but I'm less certain about that.

I'm not genuinely concerned about optimization here, just wanted to shout out that there may be some low hanging fruit :) I'll take it either way if you're not keen on that right now!

paulmelis:
Good suggestion on the DirectXMath functions. Just updated the patch and exact bounds computation time went from 99ms to 33ms. This is with the C API and a large unoptimized (85 visuals, 5.4M vertices and 5.4M indices) GLB file.

maluoi:
Wooh, that's a hefty model! Yeah, okay, I can see why that might be a good candidate for async still, I'll keep that in mind for later :) Though I'm glad to know the DXM functions helped that much!

paulmelis:
I only now noticed that this model (one of a set I've been using for months) consists of unconnected triangles :laughing: "Blame" it on modern hardware (GTX3090) and lean software like SK to make that hardly noticeable.

PR #591, opened 2023-02-10
EGL without a display server on Linux

technobaboo:
Needed for Stardust, great for many other things! This basically tries to get a display and if that fails it tries DRI from the default card, if that fails the app fails to load.

### DIFF
--- a/StereoKitC/libraries/sk_gpu.h
+++ b/StereoKitC/libraries/sk_gpu.h
@@ -541,6 +541,7 @@ typedef struct skg_platform_data_t {
 ///////////////////////////////////////////
 
 SKG_API void                skg_setup_xlib               (void *dpy, void *vi, void *fbconfig, void *drawable);
+SKG_API bool                skg_egl_dri                  ();
 SKG_API int32_t             skg_init                     (const char *app_name, void *adapter_id);
 SKG_API const char*         skg_adapter_name             ();
 SKG_API void                skg_shutdown                 ();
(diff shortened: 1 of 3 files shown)


StereoKitC/libraries/sk_gpu.h
```diff
@@ -541,6 +541,7 @@ typedef struct skg_platform_data_t {
 ///////////////////////////////////////////
 
 SKG_API void                skg_setup_xlib               (void *dpy, void *vi, void *fbconfig, void *drawable);
+SKG_API bool                skg_egl_dri                  ();
```
maluoi, line 544: For a function to be added to sk_gpu's public API, it should be implemented for both GL and D3D. This might make more sense as a capabilities check in `skg_capability` if it must be exposed?

StereoKitC/platforms/linux.cpp
```diff
@@ -438,6 +437,10 @@ bool linux_start_post_xr() {
 
 bool linux_start_flat() {
 	#if defined(SKG_LINUX_EGL)
+	if (skg_egl_dri()) {
```
maluoi, line 440: If `setup_x_window` will fail here anyhow, I think it'd be fine to let that one do its thing and error out, instead of checking on dri? I think the error there will be plenty clear.

StereoKitC/platforms/linux.cpp
```diff
@@ -1,4 +1,5 @@
 #include "linux.h"
+#include <cstddef>
```
maluoi, line 2: Codebase convention, I'd prefer this to be <stddef.h> :)

StereoKitC/platforms/linux.cpp
```diff
@@ -426,10 +421,14 @@ bool linux_start_pre_xr() {
 ///////////////////////////////////////////
 
 bool linux_start_post_xr() {
-	#if defined(SKG_LINUX_EGL)
-	if (!sk_settings.disable_desktop_input_window && !setup_x_window())
+#if defined(SKG_LINUX_EGL)
+	if (sk_settings.disable_desktop_input_window)
+		return true;
+	if(skg_egl_dri())
```
maluoi, line 427: I just don't love these calls to `skg_egl_dri`. Somewhat related to the comment below, maybe `setup_x_window` can have a mode where failure is okay, and doesn't log anything?

technobaboo:
Removed that all, made it so in XR mode the x window failing won't break
everything, and it all seems to work so that's pushed!

PR #604, opened 2023-03-14
fix: build the UWP test project with the new shader API and a C# lang…

austinbhale:
…uage (>=8.0) requirement for using @$ strings in ThemeEditor.cs


maluoi:
Thanks for catching that! :)

PR #613, opened 2023-04-02
Failed to update the dependencies with a preview Visual Studio release

austinbhale:
Working now :)


maluoi:
I actually did intentionally avoid adding the `-prerelease` flag here! My thought process was that this code is for building the StereoKit core "SDK", and should probably rely on stable build tools rather than pre-release?

Though I suppose when I move the build pipeline over to GitHub Actions, (kinda high on my current priority list) this becomes a bit of a moot point.

Alright, I convinced myself while writing this, sounds good to me. Thanks a bunch! :)

austinbhale:
Can't wait to see the automated builds! I saw the prerelease flag elsewhere in the script, so I wasn't sure if you wanted to exclude them. I'm sorta just testing the waters with the latest VS, so no prob in either case. Cheers!

PR #616, opened 2023-04-06
fix: in `World.FromPerceptionAnchor` there were times where the pose was unexpected.

austinbhale:
This returns false or an identity when converting the space into a pose fails.


maluoi:
Good fix, thank you!

PR #618, opened 2023-04-14
Renderer Screenshot overloads to get byte data via a callback

austinbhale:
WIP => still needs testing

Some of the native screenshot code is refactored to reuse similar rendering techniques. Also, the overloads were different enough to warrant new struct types: `render_screenshot_capture_t` & `render_screenshot_viewpoint_t`

Then in C#, I added a thread-safe queue to preserve the user-defined Action's lifetime until the color data is ready. Screenshots will probably always be taken synchronously but I went for thread-safety here because I can see the potential "add to list 🙅‍♂️ remove from list" error if you start taking a few screenshots in a row without receiving the first callback yet.

The part I need to double check to see is if the lifetime of the user-defined callback is preserved by being referenced by the internal callback (i.e., `RenderOnScreenshotCallback`) in the queue.

Also, all these commits definitely need squashed in the future - I had no idea that the default indentation in VS is 4 while GitHub is 8. That was pretty painful to fix lol


StereoKitC/stereokit.h
```diff
@@ -1535,6 +1535,8 @@ SK_API void                  render_add_mesh       (mesh_t mesh, material_t mate
 SK_API void                  render_add_model      (model_t model, const sk_ref(matrix) transform, color128 color_linear sk_default({1,1,1,1}), render_layer_ layer sk_default(render_layer_0));
 SK_API void                  render_blit           (tex_t to_rendertarget, material_t material);
 SK_API void                  render_screenshot     (const char *file, vec3 from_viewpt, vec3 at, int32_t width, int32_t height, float field_of_view_degrees);
+SK_API void                  render_screenshot_capture(void (*render_on_screenshot_callback)(uint8_t* out_arr_data, int32_t data_length), vec3 from_viewpt, vec3 at, int32_t width, int32_t height, float field_of_view_degrees);
+SK_API void                  render_screenshot_viewpoint(void (*render_on_screenshot_callback)(uint8_t* out_arr_data, int32_t data_length), const sk_ref(matrix) camera, const sk_ref(matrix) projection, int32_t width, int32_t height, render_layer_ layer_filter sk_default(render_layer_all), render_clear_ clear sk_default(render_clear_all));
```
maluoi, line 1539: Might also be good to make explicit that the data is rgba32 color, possibly even surface this as a `color32` instead of `uint8_t`?

StereoKitC/systems/render.cpp
```diff
@@ -86,6 +86,23 @@ struct render_screenshot_t {
 	int32_t height;
 	float   fov_degrees;
 };
+struct render_screenshot_capture_t {
```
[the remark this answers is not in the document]
maluoi: I'd probably also suggest merging `render_screenshot_viewpoint_t` and `render_screenshot_t` together too! Instead of saving directly to file, a built-in function can be automatically submitted saved in the callback that will write to file?

StereoKitC/stereokit.h
austinbhale: What is your stance on unsafe code?
I initially had this as a `Color32` type but the 'safe' way of [converting to the structure](https://github.com/StereoKit/StereoKit/blob/e72bf7d29db8b1ef79092cdf7dd8e8b851e669b5/StereoKit/Assets/Mesh.cs#L108) is unappealing imo. I looked for alternatives but a faster unsafe way is to use [Buffer.MemoryCopy](https://learn.microsoft.com/en-us/dotnet/api/system.buffer.memorycopy?view=net-8.0) which can cast the byte data to a Color32 structure.

Or do you mean, have it as a color32 type in C, but it can be a byte data type in C#?

StereoKitC/systems/render.cpp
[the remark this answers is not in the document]
austinbhale: Sweet, will do! That'll look much cleaner. That's a great idea, I'll just assign it to something that sends it off to stbi :)

StereoKitC/stereokit.h
```diff
@@ -1535,6 +1535,8 @@ SK_API void                  render_add_mesh       (mesh_t mesh, material_t mate
 SK_API void                  render_add_model      (model_t model, const sk_ref(matrix) transform, color128 color_linear sk_default({1,1,1,1}), render_layer_ layer sk_default(render_layer_0));
 SK_API void                  render_blit           (tex_t to_rendertarget, material_t material);
 SK_API void                  render_screenshot     (const char *file, vec3 from_viewpt, vec3 at, int32_t width, int32_t height, float field_of_view_degrees);
+SK_API void                  render_screenshot_capture(void (*render_on_screenshot_callback)(uint8_t* out_arr_data, int32_t data_length), vec3 from_viewpt, vec3 at, int32_t width, int32_t height, float field_of_view_degrees);
```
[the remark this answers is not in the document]
austinbhale: What is the context used for?

StereoKitC/stereokit.h
maluoi: > Or do you mean, have it as a color32 type in C, but it can be a byte data type in C#?

This is mostly what I meant here, yeah :)

SK currently has no "unsafe" code in it, but I expect that may change sometime in the future when I rework bindings for WASM? I'd probably prefer not to until that point though, I think I still had some lingering doubts about `unsafe`. Not important, but surfacing an `IntPtr` might also be a potential thing, I've seen a few libraries that would work fine with that too.

StereoKitC/stereokit.h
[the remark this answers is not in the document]
maluoi: Since C style callbacks don't "capture" the variables around them like C# callbacks do, so this is a way of passing in the "context" around that callback. For example, with a write-to-file callback, the context may include the string for the file path!

StereoKitC/stereokit.h
austinbhale: Ohh glad you brought up the `IntPtr` - I think it makes the lifetime clear. Plus it's pretty handy for testing with [this library too](https://github.com/StereoKit/StereoKit/blob/e72bf7d29db8b1ef79092cdf7dd8e8b851e669b5/StereoKit/Assets/Tex.cs#L145). Just a couple methods probably shouldn't allow for an entire project to be marked unsafe, I'm with ya on that. I've seen ridiculous performance gains just by switching to unsafe and pointer arithmetic (oh C#), but if it's for referencing some data the pointer is great

StereoKitC/stereokit.h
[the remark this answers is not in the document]
austinbhale: Ahh, so it offers that flexibility even if it's unused. Neat!

austinbhale:
Hi nick, should be ready for review. Let me know if there's anything else to change. Your prev suggestions made it a whole lot more modular, thanks!

I tested the these screenshots through the RenderCamera script by changing the `_frameSurface` TexType to something other than a `RenderTarget` and any of the following:
```
Renderer.Screenshot(
	(IntPtr data, int w, int h) => _frameSurface.SetColors(w, h, data),
	renderFrom, _renderFrom.Forward, Width, Height, fov);
```
or
```
Renderer.Screenshot(
	(IntPtr data, int w, int h) => _frameSurface.SetColors(w, h, data),
	_renderFrom.ToMatrix(), Matrix.Perspective(fov, (float)Width / Height, 0.01f, 100), Width, Height, RenderLayer.All, RenderClear.All, new Rect(500, 500, Width - 500, Height - 500));
```
or normal screenshot to a .jpg

StereoKitC/systems/render.cpp
```diff
@@ -638,60 +642,71 @@ void render_check_screenshots() {
...
 		// Create the screenshot surface
-		
 		size_t   size   = sizeof(color32) * w * h;
 		color32 *buffer = (color32*)sk_malloc(size);
-		tex_t    render_capture_surface = tex_create(tex_type_image_nomips | tex_type_rendertarget);
+
+		tex_t render_capture_surface = tex_create(tex_type_image_nomips | tex_type_rendertarget);
 		tex_set_color_arr(render_capture_surface, w, h, nullptr, 1, nullptr, 8);
-		tex_release(tex_add_zbuffer(render_capture_surface));
+		tex_t render_depth_buffer = tex_add_zbuffer(render_capture_surface);
```
maluoi, line 651: This is perfectly fine and doesn't need changed, I just thought I'd note that the previous way was also correct! When adding a zbuffer to a texture, that texture holds a reference to the zbuffer internally to ensure the zbuffer's lifetime matches that of the parent texture.

StereoKit/Systems/Renderer.cs
```diff
@@ -284,6 +291,73 @@ public static void Screenshot(Vec3 from, Vec3 at, int width, int height, string
...
+		/// <param name="at">Direction the viewpoint is looking at.</param>
+		/// <param name="width">Size of the screenshot horizontally, in pixels.
+		/// </param>
+		/// <param name="height">Size of the screenshot vertically, in pixels.
+		/// </param>
+		/// <param name="fieldOfViewDegrees">The angle of the viewport, in 
+		/// degrees.</param>
+		public static void Screenshot(ScreenshotCallback onScreenshot, Vec3 from, Vec3 at, int width, int height, float fieldOfViewDegrees = 90)
+		{
+			if (_renderCaptureCallbacks is null) _renderCaptureCallbacks = new ConcurrentQueue<RenderOnScreenshotCallback>();
```
maluoi, line 314: I hadn't actually seen the matching form of the `is` keyword before, no action needed here, but I'm glad you've brought it to my attention because I rather like how it works!

StereoKit/Systems/Renderer.cs
```diff
@@ -284,6 +291,73 @@ public static void Screenshot(Vec3 from, Vec3 at, int width, int height, string
...
+		/// surface!</param>
+		public static void Screenshot(ScreenshotCallback onScreenshot, Matrix camera, Matrix projection, int width, int height, RenderLayer layerFilter = RenderLayer.All, RenderClear clear = RenderClear.All, Rect viewport = default(Rect))
+		{
+			if (_renderCaptureCallbacks is null) _renderCaptureCallbacks = new ConcurrentQueue<RenderOnScreenshotCallback>();
+			RenderOnScreenshotCallback renderCaptureCallback = (IntPtr dataPtr, int w, int h, IntPtr context) =>
+			{
+				if (_renderCaptureCallbacks.TryDequeue(out _))
+					onScreenshot.Invoke(dataPtr, w, h);
+			};
+			_renderCaptureCallbacks.Enqueue(renderCaptureCallback);
```
maluoi, line 357: This enqueue dequeue thing feels a little shaky to me! The only thing that holds this to be correct is that both the `ConcurrentQueue` and the underlying processing of screenshots are independently FIFO. I think it's still the right data structure, but it might be good to note that this is the case in the comments of `render_check_screenshots`, so if we change the behavior of callbacks in the future to be more async, we know to still preserve the FIFO behavior of the code.

maluoi (commented):
Really nice work here as always! A few more really small things, and I'll merge this in :)

StereoKit/Systems/Renderer.cs
austinbhale: I really like them too - they're readable and make the ReferenceEquals intention clear. Also, when you update to a newer C# language version, you can later null-coalescing assignment ??= (which also uses the "is" check) for this line too :)

StereoKitC/systems/render.cpp
austinbhale: I’m glad its lifetime matches, sweet! I’ll def change it back. I also saw you noted the return of the reference is getting removed in v0.4

StereoKitC/systems/render.cpp
```diff
@@ -878,9 +893,34 @@ void render_blit(tex_t to, material_t material) {
...
+
 void render_screenshot(const char *file, vec3 from_viewpt, vec3 at, int width, int height, float fov_degrees) {
 	char *file_copy = string_copy(file);
-	render_screenshot_list.add( render_screenshot_t{ file_copy, from_viewpt, at, width, height, fov_degrees });
+	matrix view = matrix_trs(
+		from_viewpt,
+		quat_lookat(from_viewpt, at));
+	matrix_inverse(view, view);
+	matrix proj = matrix_perspective(fov_degrees, (float)width / height, render_clip_planes.x, render_clip_planes.y);
+	render_screenshot_list.add( render_screenshot_t{ render_save_to_file, file_copy, view, proj, rect_t(), width, height, render_layer_all, render_clear_all });
```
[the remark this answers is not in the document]
austinbhale: Definitely not nit picky! I just read about how its ambiguity and how it may even lead to the "most vexing parse" :o plus it's recommended all t/o the [docs](https://learn.microsoft.com/en-us/cpp/cpp/initializing-classes-and-structs-without-constructors-cpp?view=msvc-170). Thanks for bringing this up!

StereoKit/Systems/Renderer.cs
austinbhale: Having an async callback sounds interesting. It sorta just fires them all natively and the enqueued user-defined callbacks can be done in any order. Hmm maybe a list with a lock is the way to go after all, in order to finalize the completed actions

But without async callbacks, it's pretty much FIFO. I guess it could change from ConcurrentQueue to Queue since it's not actually thread safe, as the current C# enqueue and native C list add are mutually exclusive without a lock

austinbhale: Oh! What if we pass in another callback natively that essentially says "# of screenshots are complete" and this can be used by C# to dequeue that many times, or any C API wanting to know how many of the screenshots were just processed?

maluoi: I think that even if this becomes more async, it will still likely be FIFO? I believe the GPU should still do the work in order, so it shouldn't be much of a burden to just ensure that actually happens and promise it to the developer. Having that promise seems valuable anyhow, like if I wanted to screenshot a series of images for a gif, having them arrive out-of-order would just complicate matters.

But yeah, I guess the native list isn't threadsafe, so the C# side being safe there doesn't make it much better! But `ConcurrentQueue` _shouldn't_ add much overhead, so feel free to leave it in! From what I've read, it's implemented with atomic operations rather than locks. I've not so long ago figured how to do some atomic stuff in C, so at some point later on I'd like to revise SK's builtin data structures to utilize that.

maluoi: I've thought about having a poll based alternative for anything that uses callbacks? I think that'd probably be a lot more "StereoKit idiomatic", but determining the source of things in a poll queue might be a little problematic here! I don't think I'm quite going to do this though.

austinbhale: I changed it back to a `Queue` mainly so I remember to call the screenshot from SK's main thread (via a stepper or `SK.ExecuteOnMain`). Then tested it from a separate thread and it worked great without the added concurrency. [I don't think a ConcurrentQueue is too relevant for this scenario](https://learn.microsoft.com/en-us/dotnet/standard/collections/thread-safe/when-to-use-a-thread-safe-collection#concurrentqueuet-vs-queuet), with the FIFO order guaranteed. I'll also add a FIFO comment in there then ready for review again.

I guess it'd be like polling for OpenXR events but SK style.

austinbhale: > Having that promise seems valuable anyhow, like if I wanted to screenshot a series of images for a gif, having them arrive out-of-order would just complicate matters.

Huh now I want a script to make speedy gifs for sharing! :D

maluoi: That was kinda the idea with the RenderCamera script! I just use ScreenToGif for converting the bmp frames into the final gif :) I'm sure there's a NuGet out there that could handle gif stuff instead, but I didn't want to worry about that for a single file tool. I had been thinking about adding more direct gif support at one point ([this library](https://github.com/notnullnotvoid/msf_gif) in particular looked super nice), and now that I'm peeking back at it I'm not too sure why I didn't!

maluoi:
Fantastic, thank you so much! :)

StereoKit/Systems/Renderer.cs
austinbhale: I'm a big fan of changing the perspectives in RenderCamera. Woa msf_gif looks so clean and writes a frame at a time to spare the memory! A few of their [comments](https://github.com/notnullnotvoid/msf_gif/blob/7829c8f439d013deeb54eba94761403e1da2a960/msf_gif.h#L318) made me chuckle. As a single header with RGBA format by default, this could be a real winner

PR #622, opened 2023-04-21
Renderer Screenshot with an adjustable texture format

austinbhale:
Hey nick, turns out I need a different format than RGBA, so I just added a parameter for changing the texture format. Worked as expected in the `RenderCamera` script.

Does it need support for all of the texture types? I just need it for BGRA which is the same buffer size. Maybe the others could be marked as unsupported for the time being?


maluoi:
This looks like an improvement to me! BGRA does kinda raise an interesting question though, what should happen if the user requests a screenshot in a format that's not valid for render textures? BGRA is not, afaik, universally supported for render target surfaces. Mostly on different platforms.

austinbhale:
If you create a render target with an unsupported format, is [this](https://github.com/StereoKit/StereoKit/blob/aae289a27ad8275466da31e74a66ecf9814889a5/StereoKitC/libraries/sk_gpu.h#L2132) where it'd fail (or earlier)?

It turns out I don't need BGRA anymore, but would be happy to add some error handling in here for unsupported pixel formats!

maluoi:
I would expect [this spot](https://github.com/StereoKit/StereoKit/blob/aae289a27ad8275466da31e74a66ecf9814889a5/StereoKitC/libraries/sk_gpu.h#L2222) to be where the failure starts, that function call would have enough information to decide if it was a valid texture setup or not! But it does let execution fall through there, so it would probably halt and return a bit later when trying to create the first view item.

austinbhale:
If the current error handling is a critical log, could we keep the same behavior here? Otherwise we'd have to change the return type of `skg_tex_set_contents_arr` to a bool or throw

maluoi:
Yeah, you're right, I think there's probably some work to be done here related to error reporting. But I don't think it's in need of immediate attention here, and can probably be a bit of a separate project later on. The logs should already highlight the core issues if they crop up!

PR #627, opened 2023-05-05
Override the sprite material with a custom one

austinbhale:


maluoi:
One thing that makes me a little wary here is that an Atlas can be shared between many `Sprite`s, and so setting the `Material` of a `Sprite` on an existing Atlas would change the `Material` for the whole Atlas. This is not exactly a problem for `Single` type sprites, but does make the situation overall much stickier.

And yeah, I realize Atlas sprites technically aren't working yet, but I do have them almost finished in a branch somewhere 😅 

I think the _correct_ solution here would be to formalize an `Atlas` as a proper StereoKit `Asset`, and have more direct control over the `Atlas`'s settings and `Material`s? This way you could pass in the `Atlas` the `Sprite` belongs on, pre-configured with the correct `Material`. But this is a bigger task to chew on 😬 and might be better to do over in the Atlasing branch.

As a more immediate solution, I'd probably prefer having a `sprite_get_atlas_material()`/`sprite.AtlasMaterial` get property that would allow you to modify the `Material` after creation? I think this is probably a nicer, less impactful thing to either leave in, or deprecate later when Atlas assets are implemented?

austinbhale:
Thanks for the update on them! I'm curious to see how they'll be batched :)

I agree that an `Atlas` would serve well as an `Asset`. So would an atlas have a list of associated sprites and every sprite an associated atlas? And even if a sprite is a `Single`, could it have a dummy `Atlas` property that would only reference its material? Trying to think of a way where you wouldn't have to pass in the atlas, but this may couple them a bit

Hadn't thought of replacing a material's shader, that's pretty nifty

I don't think this PR is more pressing than getting the core of the atlases right. I can def work around sprites currently. Cheers!

PR #632, opened 2023-05-10
Bytedance Pico OpenXR Loader, Controller Profiles and Handtracking

rayperkins:
Changes have been made to support the Pico 4 headset.

The standard/Oculus libopenxr_loader.so does not work on the Pico, its behaviour is to launch into a black display with the loading indicator.  Instead, Pico provide their own loader. The changes are to include it into the Nuget package and allow a developer to set `<SKOpenXRLoader>Pico</SKOpenXRLoader>` in the csproj file and have the Pico specific loader bundled in the app. The Oculus loader is still used during the xmake build, there's an assumption here that api surface is the same between the Pico and Oculus loader.

I have not been able to get the whole project to build in order to test the nuget packing works correctly. 

The controller profiles are added for the Pico 4 and Pico Neo 3.  Note that the xrGetCurrentInteractionProfile returns '/interaction_profiles/bytedance/pico_neo3_controller' instead of the expected '/interaction_profiles/bytedance/pico4_controller'. I don't think this is a StereoKit issue and the controllers appear to work OK.

Hand tracking was not working on the Pico due the joint at index 0 (palm center) being empty and so failing a valid check. The change to resolve this is to use the wrist joint (index 1). However, this change assumes that existing devices will always present a wrist joint. I can only confirm this is the case on the Pico 4 device.

Features confirmed working on the Pico 4 OS 5.5.0
- Pico's version of the loader.
- Pico 4 Controller with adjusted orientations.
- Hand tracking
- StereoKitTest UI demo tested
- Pass Through. Note that disabling pass through from within StereoKitTest causes a crash. Have yet to debug but suspect the crash is in the next FrameEnd cycle after the pass through is disabled.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -31,6 +31,11 @@ In addition to the Windows steps, to build for Android, StereoKit uses [xmake](h
 
 I may update this guide eventually, but if you can track what steps you take while doing this and submit a PR, that would be really incredible!
 
+General Setup steps:
+1. Install xmake
+2. Install the Android NDK. Grab the latest version from [here](https://developer.android.com/ndk/downloads) and unzip into a folder, a path without spaces will make life easier (C:\tools\Android\android-sdk)
+3. Set the NDK path into xmake. This command sets the global config: `xmake g --ndk="C:\tools\Android\android-sdk\android-ndk-rXYZ"`
+
 ## I want to modify code (Linux)
 
 StereoKit official Linux builds currently use xmake running on Windows Subsystems for Linux, however, it's also possible to use cmake, and do either from a normal Linux machine.
--- a/Examples/StereoKitTest/StereoKitTest_Maui/Platforms/Android/AndroidManifest.xml
+++ b/Examples/StereoKitTest/StereoKitTest_Maui/Platforms/Android/AndroidManifest.xml
@@ -3,9 +3,12 @@
 	<uses-sdk android:minSdkVersion="29" android:targetSdkVersion="29" />
 
 	<application android:allowBackup="true" android:icon="@mipmap/appicon" android:roundIcon="@mipmap/appicon_round" android:supportsRtl="true">
+		<!-- Oculus -->
 		<meta-data android:name="com.oculus.supportedDevices" android:value="quest|quest2|questpro"/>
 		<meta-data android:name="com.oculus.handtracking.version" android:value="V2.0"/>
-		<meta-data android:name="pvr.app.type" android:value="vr" /> <!-- Pico -->
+		<!-- Pico -->
+		<meta-data android:name="pvr.app.type" android:value="vr" />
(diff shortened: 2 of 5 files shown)


maluoi:
One thing I'd like to avoid here is adding the Pico `openxr_loader.so` to the repository! I know I've done this in the past, but I was kinda new to native development at the start of this project. It bloats the repository size and isn't a terribly great practice, especially now that Android runtimes are starting to move on from custom loaders! I actually need to remove all the Oculus stuff, since Oculus now properly supports the standard loader as of their latest release.

I wouldn't really worry about adding anything to the xmake file, the `oculus-openxr` config setting is an older name, and the most important thing it's really doing is making sure the standard loader isn't static linked, iirc?

>  The change to resolve this is to use the wrist joint (index 1). However, this change assumes that existing devices will always present a wrist joint. I can only confirm this is the case on the Pico 4 device.

I might suggest the index finger tip! I think that's probably the joint most likely to be available in general :)

maluoi:
Another thought, if you re-arrange the pico4 profile to be before the pico3 profile, will the runtime then select that profile instead?

rayperkins:
It does make sense to keep SK agnostic of the loader, especially that an application has to be targeted for the various platforms. Any required libs can be included per platform when an app gets packaged. Perhaps I can update the docs with a write up to get the Pico running using SK.

I've pulled out the loader lib for the pico and have left the oculus stuff there as I haven't the hardware to test changes on.
However, I get hold of a Quest 2 for testing and make another PR for those changes.

I went with the wrist as I figured the visual system would always be able to track it if the hand is visible and that maybe a closed fist would make finger tracking stop working. But, this is a wild guess as I really don't know what magic goes into the tracking system. Anyway, the fingertip sense as it's the most important joint needed for interactions. I've update the code to suit.

I had a go at changing the profile ordering. It only loaded the pico 4 profile when the pico 3 neo was commented out. The [spec says](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#input-suggested-bindings) the XR runtime may select one of the suggested interaction profiles. It seems that the Pico 4 prefers the neo 3 profile.

maluoi:
Perfect, this looks great to me, thanks for the changes!

I'm a little curious about the occlusion case, and if the finger might "untrack". I did think about that a bit, but I don't believe that's how hand tracking systems currently behave? They either see the hand, hallucinate the hand, or don't give you anything at all! But I'll keep an eye out for this as devices move forward, I really wish there was just a better mechanism for this overall, and I'm _hoping_ the next iteration of hand tracking extensions might account for some of this!

maluoi:
Oh, actually, did this all work even without asking for the `XR_BD_controller_interaction` extension?

rayperkins:
I didn't know it was needed and so didn't explicitly request the feature. However, the controllers seemed to work just fine.

PR #637, opened 2023-05-22
Overload for off-center perspective matrix using camera's principal point

austinbhale:
This overload is for precise rendering from a camera sensor's undistorted view. (+X right, +Y up) stays consistent with the game engine's coordinate system. The left and bottom are negative values from the camera's center, while the right and top are positive values.

### DIFF
--- a/StereoKit/Math/Matrix.cs
+++ b/StereoKit/Math/Matrix.cs
@@ -465,8 +465,8 @@ namespace StereoKit
 		{
 			float l = nearClip / focalLength.x * -principalPoint.x;
 			float r = nearClip / focalLength.x * (imageResolution.x - principalPoint.x);
-			float t = nearClip / focalLength.y * principalPoint.y;
 			float b = nearClip / focalLength.y * (principalPoint.y - imageResolution.y);
+			float t = nearClip / focalLength.y * principalPoint.y;
 			return Matrix4x4.CreatePerspectiveOffCenter(l, r, b, t, nearClip, farClip);
 		}
 
--- a/StereoKit/Math/Matrix.cs
+++ b/StereoKit/Math/Matrix.cs
@@ -445,6 +445,32 @@ namespace StereoKit
 		}
 		
 		/// <summary>This creates a matrix used for projecting 3D geometry
+		/// onto a 2D surface for rasterization. With the known camera 
+		/// intrinsics, you can replicate its perspective!</summary>
+		/// <param name="imageResolution">The resolution of the image. This
+		/// should be the image's width and height in pixels.</param>
+		/// <param name="focalLength">The focal length of the camera in pixels,
+		/// with image coordinates +X (pointing right) and +Y (pointing up).</param>
+		/// <param name="principalPoint">The principal point of the camera in pixels,
+		/// with image coordinates +X (pointing right) and +Y (pointing up).</param>
+		/// <param name="nearClip">Anything closer than this distance (in
+		/// meters) will be discarded. Must not be zero, and if you make this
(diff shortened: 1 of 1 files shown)


StereoKit/Math/Matrix.cs
```diff
@@ -444,6 +444,32 @@ public static Matrix Perspective(Vec2 imageResolution, Vec2 focalLength, float n
...
 		}
 		
+		/// <summary>This creates a matrix used for projecting 3D geometry
+		/// onto a 2D surface for rasterization. With the known camera 
+		/// intrinsics, you can replicate its perspective!</summary>
+		/// <param name="imageResolution">The resolution of the image. This
+		/// should be the image's width and height in pixels.</param>
+		/// <param name="focalLength">The focal length of the camera in pixels,
+		/// with image coordinates +X (pointing right) and +Y (pointing up).</param>
+		/// <param name="principalPoint">The principal point of the camera in pixels,
```
maluoi, line 454: What is a principal point? I'm a little fuzzy on the idea.

StereoKit/Math/Matrix.cs
```diff
@@ -444,6 +444,32 @@ public static Matrix Perspective(Vec2 imageResolution, Vec2 focalLength, float n
 			return Matrix4x4.CreatePerspective(nearPlaneDimensions.x, nearPlaneDimensions.y, nearClip, farClip);
 		}
 		
+		/// <summary>This creates a matrix used for projecting 3D geometry
+		/// onto a 2D surface for rasterization. With the known camera 
+		/// intrinsics, you can replicate its perspective!</summary>
+		/// <param name="imageResolution">The resolution of the image. This
+		/// should be the image's width and height in pixels.</param>
+		/// <param name="focalLength">The focal length of the camera in pixels,
```
maluoi, line 452: I'm also a little uncertain what focal length in pixels means, could you elaborate on this a bit? I would understand degrees or millimeters here, but pixels is a bit baffling!

austinbhale: Ohh yep, it's in millimeters. I copied the mistake from the other comment so I'll update both

maluoi: Millimeters also has some wiggle room, it's usually dependent on sensor size, so I guess it may also be nice to clarify that a bit too then?

austinbhale: Sorry just double checked - actually it is in pixels so the original description is correct. The focal length should be in the same unit as the image resolution for `Vec2 nearPlaneDimensions = imageResolution / focalLength * nearClip;` to work. In OpenCV as well, the camera matrix includes a focal length and principal point in pixel units. You only need to know the sensor size if you convert from pixels -> mm & vice versa

StereoKit/Math/Matrix.cs
austinbhale: It's usually in the center of the image, but for some cases, there's an offset from where the line (i.e., optical axis) going straight out from the camera hits the image plane

austinbhale:
Should I include some of these clarifications in the remarks sections?

austinbhale:
Just added comments in the remarks, hopefully that clears things up

> For units, those should generally go in the function parameter's name
Do you mean change the variable name to `focalLengthPixels` or leave it how it's described in the docs?

maluoi:
I'd love it if it could be `focalLengthPixels` or `focalLengthPx`, yeah :)

austinbhale:
Good idea! That makes the calculations pretty clear. I didn't rename `imageResolution` since that seemed a bit redundant

maluoi:
Perfect, thank you!

PR #641, opened 2023-05-28
fix: eek! accidentally added a tex format parameter for RenderTo from…

austinbhale:
… #622 which C# can't find in DLL

and here I thought I was having submodule issues 🤦


maluoi:
Oops, thanks for catching that one! I'm really looking forward to pushing some CI stuff here to help prevent that sort of thing :D

PR #644, opened 2023-06-08
Fix mistake in docs

Strepto:
Fixes a minor mistake in the docs where the result type should be in local space, but the docs say World Space


maluoi:
Thanks for catching that one!

PR #646, opened 2023-06-10
fix: dlopen & dlopen2 arguments interpreted as absolute unless clearly specified as relative paths

austinbhale:
Here's a sample script to test with and without './'

```
using System;
using System.Runtime.InteropServices;

public class Program
{
    [DllImport("libdl.so")]
    public static extern IntPtr dlopen(string fileName, int flags);

    [DllImport("libdl.so")]
    public static extern IntPtr dlerror();

    public static void Main(string[] args)
    {
        IntPtr handle = dlopen("./libStereoKitC.so", 2);
        if (handle == IntPtr.Zero)
        {
            IntPtr error = dlerror();
            string message = Marshal.PtrToStringAnsi(error);
            Console.WriteLine($"Error: {message}");
        }
        else
        {
            Console.WriteLine("Library loaded successfully.");
        }
    }
}
```

### DIFF
--- a/StereoKit/Native/NativeLib.cs
+++ b/StereoKit/Native/NativeLib.cs
@@ -55,7 +55,7 @@ namespace StereoKit
 			const int RTLD_NOW = 2;
 			if (dlopen($"./runtimes/linux-{arch}/native/libStereoKitC.so", RTLD_NOW) != IntPtr.Zero) return true;
 			if (dlopen($"{AppDomain.CurrentDomain.BaseDirectory}/runtimes/linux-{arch}/native/libStereoKitC.so", RTLD_NOW) != IntPtr.Zero) return true;
-			if (dlopen("libStereoKitC.so", RTLD_NOW) != IntPtr.Zero) return true;
+			if (dlopen("./libStereoKitC.so", RTLD_NOW) != IntPtr.Zero) return true;
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


austinbhale:
Oh this check is seeing if it's on your path. That makes sense! Closing this then

maluoi:
Yeah, that code is indeed for checking the path, or whatever Linux's equivalent is! libStereoKitC.so shouldn't really be in the project root next to the executable, since the same core executable can work on multiple architectures, and the runtimes folder contains StereoKit binaries for those architectures! I think `netstandard2.1` has some additional tools that can help a lot here, but I've been unable to easily adopt those since UWP is locked to `netstandard2.0`.

I don't know if maybe it could be worth adding in a check for `./libStereoKitC.so` before the `libStereoKitC.so` check, just in case people want to ship a lighter weight or simpler package that supports only one architecture? I'm also surprised if `libStereoKitC.so` wouldn't check the root folder anyhow, but Linux has been anything but predictable so far here.

austinbhale:
Oh man being locked to netstandard2.0 has been sad but luckily not impossible to workaround yet.

Tbh I’m not sure if it's a huge priority. I feel like if they want to support one architecture it's not the worst thing to have the extended path, otherwise I’d vote for having it in the AppDomain base directory so it's always in the proper Debug/Release sections

PR #655, opened 2023-07-03
fix: crashing `mouse_scroll` call from HoloLens 2 UWP

austinbhale:
Hey nick, so happy the newest updates came through! I tested this *early adopter* version on the HL2 and receive this native exception at the start (& every step after). Is a simple try/catch enough to handle this call or do you want to check for mouse availability elsewhere?

### DIFF
--- a/StereoKitC/platforms/uwp.cpp
+++ b/StereoKitC/platforms/uwp.cpp
@@ -436,9 +436,7 @@ void uwp_set_mouse(vec2 window_pos) {
 ///////////////////////////////////////////
 
 float uwp_get_scroll() {
-	if (ViewProvider::inst != nullptr) {
-		return ViewProvider::inst->mouse_scroll;
-	}
+	return ViewProvider::inst != nullptr ? ViewProvider::inst->mouse_scroll : 0;
 }
 
 ///////////////////////////////////////////
--- a/StereoKitC/platforms/uwp.cpp
+++ b/StereoKitC/platforms/uwp.cpp
@@ -436,11 +436,8 @@ void uwp_set_mouse(vec2 window_pos) {
 ///////////////////////////////////////////
 
 float uwp_get_scroll() {
-	try {
+	if (ViewProvider::inst != nullptr) {
 		return ViewProvider::inst->mouse_scroll;
-	}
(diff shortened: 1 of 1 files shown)


maluoi:
Fast work there! This doesn't appear to happen in my local build that I can see. I assume this is just a null reference exception? In which case, checking for nullptr on `inst` beforehand would be preferable to a try/catch. But the more interesting item would be why it's null in the first place! That's probably an indicator of a larger failure earlier with creating the window or something, preventing `inst` from getting set to begin with.

Unless this is something else?

austinbhale:
Yep, you're right in that it's a nullptr. I looked for every point where it creates the ViewProvider instance, and I believe it's only created for flatscreen mode in SK. Do you only handle the UWP dispatcher/core window events for non-immersive apps, or is that HL2 exclusive? If it's only for XAML apps, why call `input_mouse_update` for all UWP targets if it relies on a ViewProvider instance?

maluoi:
Oh wow, yeah looks like you're right! `input_mouse_update` calls down to this while there should be no instance. I guess this may be one of those "how did that ever work??" moments! A null check is probably quite appropriate here.

Right now, mouse only really works in a simulator context. I still need to do the work to make it a useful thing in XR.

austinbhale:
Sure is a mystery, maybe with some different build flags or optimizations thrown in, or that it's running on Windows 11 now? This seems like a good crash then :)

I added the null check in, lmk if you were thinking something else. Mouse input in XR could be like an inverse remote assist heh

maluoi:
Oh... "not all control paths return a value" is only a warning? Wild! There's no return value there in the case of null :)

This is where I like to use a more Functional approach and use a ternary operator. Regular conditionals used for setting a value can be a little dangerous due to missed paths like this.

austinbhale:
Total oversight! Sorry about that. I need to enable that error message natively, C feels dangerous if not

maluoi:
Either way, thanks for catching this one! Not sure why that never manifested on my computer, I may be overdue for reformatting this computer or something.

austinbhale:
You have everything automated enough you could easily get it back :D
Congrats on the release(s)!

Issue #486, opened 2022-10-11
Support for different keyboards


spatialfree:
Make flatscreen simulation input accessible to a host of custom keyboards/layouts and languages. By mapping the keys based on their Physical Location instead of Value.

I've noticed that Unity's implemented this in their newer [input system](https://forum.unity.com/threads/new-input-system-which-keyboard-layouts-are-supported.912074/). As well as a variety of games/apps seem to work this way as well, but I can't seem to find the common thread just yet... One easy way to test for this is to install the [colemak](https://colemak.com/) layout and toggle it on-off.

maluoi:
Could you elaborate on this a little more? StereoKit does support Unicode input (try changing your keyboard to say, Cyrillic), and input text into a `UI.Input`. If you're referring to `Input.Key`, these map directly to [virtual keys](https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes) which are different from text input.

If you're looking to read text input rather than keys, you can access these through [`Input.TextConsume`](https://stereokit.net/Pages/StereoKit/Input/TextConsume.html)/`TextReset`!

spatialfree:
Thanks! That link helped me find the correct relevant term: "[Scan Codes](https://learn.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input#scan-codes)"

maluoi:
Ah ok, does this mean you're wanting the current `Key`s to behave less like virtual keys, and more like scan codes?

spatialfree:
Yeah! For when you are flying around on desktop, the key position matters more than the "value".

PR #657, opened 2023-07-06
keycodes

spatialfree:
i took a stab at an issue i posted a while back: https://github.com/StereoKit/StereoKit/issues/486

and the following commit actually manages to do what i need for linux ^-^ ...but windows scan codes are different than x11's keycodes~ so that will need to be handled properly before merging these changes!


maluoi:
Heyas, thanks for the PR here! I'm still thinking about this, but the shape of the API changes here don't feel right to me. This is mixing together the scan codes with the virtual key codes, when they're two rather separate ideas. Like, injecting a scan code should automatically inject the matching virtual key, and injecting a virtual key should automatically inject a scan code as well. Testing for a key should likewise be separate from testing for a scan code.

I'm also not terribly excited about using numbers for scan codes? While I understand that scan codes may not have good mappings to real names, it still feels like a really bad experience. Perhaps scan codes can be numbers, _or_ an enumeration that maps QWERTY to scan codes?

On the technical size, a `uint8_t` would match a `byte` in C#.

spatialfree:
thanks for being so thorough! as it really helped me iterate on this~ a whole lot quicker! :+1: 

i made what seemed like some good tradeoffs of complexity clarity and ease of use inline with your feedback while also making it easier to plug in the windows platform side of things

PR #662, opened 2023-07-07
Fix an improper return from openxr_poll_events, and add a poll with a…

maluoi:
Fix an improper return from openxr_poll_events, and add a poll with a potential "out" for the get_app_space loop.


moshimeow:
LGTM

PR #665, opened 2023-07-11
Don't blindly close the X display handle on linux shutdown

pelrun:
linux_shutdown() segfaults by trying to do a XCloseDisplay(NULL) in the cases where setup_x_window() wasn't called during init.


maluoi:
Ty for this!

PR #713, opened 2023-09-01
Cmake Support Linux Arm64 builds. 

### LINKED ISSUE #702: Linux ARM64 Builds in CI
## Issue

Linux ARM64 builds were temporarily removed from the NuGet builds after v0.3.7-preview.9 in the switch from xmake to cmake. These need re-added before v0.3.7 can be pushed out. Unfortunately, Linux ARM64 is _much_ trickier to build with cmake than with xmake, and I still haven't figured out how exactly to do this.

## Current State (Linux x64)

Currently, the build for Linux x64 looks approximately like this:
```shell
sudo apt-get update
sudo apt-get install lld libfontconfig1-dev libgl1-mesa-dev libx11-xcb-dev libxfixes-dev libxcb-dri2-0-dev libxcb-glx0-dev libxcb-icccm4-dev libxcb-keysyms1-dev libxcb-randr0-dev libxrandr-dev libxxf86vm-dev mesa-common-dev

cmake --workflow --preset Linuxx64Release

# Or

cmake --preset Linuxx64Release
cmake --build --preset Linuxx64Release
```

Where the preset configure looks like this ([abbreviated](https://github.com/StereoKit/StereoKit/blob/f5ff74818bbe8bf607de8284cc86c2439837871c/CMakePresets.json#L135-L164)):
```json
{
	"name": "Linux",
	"generator": "Ninja",
	"cacheVariables": {
		"CMAKE_SYSTEM_NAME": "Linux",
		"CMAKE_C_COMPILER": "clang",
		"CMAKE_CXX_COMPILER": "clang++",
		"CMAKE_LINKER": "lld",
		"CMAKE_EXE_LINKER_FLAGS": "-fuse-ld=lld",
		"CMAKE_SHARED_LINKER_FLAGS": "-fuse-ld=lld",
		"CMAKE_MODULE_LINKER_FLAGS": "-fuse-ld=lld"
	},
},
{
	"name": "Linuxx64Release",
	"inherits": "Linux",
	"binaryDir": "${sourceDir}/bin/intermediate/cmake/Linux_x64_Release",
	"cacheVariables": {
		"CMAKE_BUILD_TYPE": "Release"
	}
},
```

## Linux ARM64

While it's [not too hard](https://github.com/StereoKit/StereoKit/blob/develop/CMakePresets.json#L184-L202) to set up some flags to tell cmake to cross-compile to ARM64, the difficulty is with the `apt-get` dependencies that need installed. By default, `apt-get` only installs packages for the local build machine's architecture (x64), and some amount of setup must be done to install ARM64 flavors of the same libraries.

Unfortunately, I've had mixed luck with setting up `apt-get` for this. In theory, it should look [something like this](https://askubuntu.com/questions/430705/how-to-use-apt-get-to-download-multi-arch-library):

```shell
dpkg --add-architecture arm64

# Add arm64 package source URLs to sources.list
# /etc/apt/sources.list
# Something like this?
# deb [arch=arm64] http://ports.ubuntu.com/ jammy universe

sudo apt-get update
sudo apt-get install libfontconfig1-dev:arm64 ...
```

The last few times I tried this, I ended up breaking `apt-get`, probably with bad edits to the `sources.list` file. Also having not gotten past this point, there may need to be some additional edits to the `CMakePresets.json` to take advantage of the arm64 libraries, possibly at least by pointing to the folder where the arm64 libraries reside?

## Moving Forward

The paths forward that I see are:

- Finding the correct modifications to the `sources.list` file, and figuring out how to fit that all into the CI pipeline.
- Emulate a Linux Arm64 machine via something like docker, qemu, or whatever.
- Figure out how to cross-compile with Zig, though it may still have all the same problems, but with musl in the mix.

ponugotimanasaqc:
Added build in docker container as installing arm64 packages overwrites lld which fails the x64 build.

How the docker build works : 
1. Creates a docker image  with all the necessary packages installed. 
2. Mounts the source code in docker container.
3. Runs the CMake command to build.

How to run without docker : 
1. Setup cross compiling for ARM. Requires updating repos and installing arm packages. See docker/arm64linux.Dockerfile for the steps.
2. cmake --preset Linux_x64_Release
3. cmake --build Linux_x64_Release

How to run using docker : 
1. cd docker
4. ./build.sh

FIxes #702

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -203,15 +203,18 @@
       "binaryDir":   "${sourceDir}/bin/intermediate/cmake/Linux_ARM64_Release",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE":       "Release",
-        "CMAKE_AR":               "aarch64-linux-gnu-ar",
-        "CMAKE_LINKER":           "aarch64-linux-gnu-ld",
-        "CMAKE_OBJCOPY":          "aarch64-linux-gnu-objcopy",
-        "CMAKE_RANLIB":           "aarch64-linux-gnu-ranlib",
-        "CMAKE_SIZE":             "aarch64-linux-gnu-size",
-        "CMAKE_STRIP":            "aarch64-linux-gnu-strip",
+        "CMAKE_AR":               "/usr/bin/aarch64-linux-gnu-ar",
+        "CMAKE_LINKER":           "/usr/bin/aarch64-linux-gnu-ld",
+        "CMAKE_OBJCOPY":          "/usr/bin/aarch64-linux-gnu-objcopy",
+        "CMAKE_RANLIB":           "/usr/bin/aarch64-linux-gnu-ranlib",
+        "CMAKE_SIZE":             "/usr/bin/aarch64-linux-gnu-size",
+        "CMAKE_STRIP":            "/usr/bin/aarch64-linux-gnu-strip",
         "CMAKE_C_COMPILER":       "aarch64-linux-gnu-gcc",
         "CMAKE_CXX_COMPILER":     "aarch64-linux-gnu-g++",
-        "CMAKE_SYSTEM_PROCESSOR": "arm64"
+        "CMAKE_SYSTEM_PROCESSOR": "arm64",
+        "CMAKE_EXE_LINKER_FLAGS": null,
+        "CMAKE_SHARED_LINKER_FLAGS": null,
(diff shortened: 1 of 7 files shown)


maluoi:
This looks pretty good for now! I do have some thoughts about Ubuntu 22.04 vs 20.04, but I think you might have to get into a bunch of other stuff to do that? If that's an easy change for this, that might be nice as well.

PR #816, opened 2023-12-22
fix: divide by 0 exception for views with no color swapchain

austinbhale:
Hi Nick! Just updated SK dev branch after a few months and ran into a native divide by 0 exception for the HL2's secondary view. I left more details in the commit. Lmk if you'd rather address this another way instead, thanks :)

### DIFF
--- a/Examples/StereoKitTest/Program.cs
+++ b/Examples/StereoKitTest/Program.cs
@@ -96,9 +96,7 @@ class Program
 
 		Init();
 
-		SK.Run(Step);
-
-		Tests.Shutdown();
+		SK.Run(Step, Tests.Shutdown);
 	}
 
 	static void Init()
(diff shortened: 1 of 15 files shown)


maluoi:
Hey Austin! Thanks for pointing this one out :) It's been a while since I've gotten to see the secondary view path running first-hand, guess I botched this bit while cleaning things up!

Just for a quick refresher, I guess the secondary view reports a view count of 0 from `xrEnumerateViewConfigurationViews` when recording isn't in progress? If that's the case, I might want to just double check a bunch of that code to make sure it handles that correctly. Your code here might be sufficient, but I may wanna just spend a bit of extra time running through it a bit myself too.

Also, a change to how the OSS stuff on this repo works, slightly more formal these days :) You will need to add a [DCO (Developer Certificate of Origin)](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin) sign-off line to your commits, or add a remediation commit for it (details here: https://github.com/StereoKit/StereoKit/pull/816/checks?check_run_id=19905206975). And preferably stick your name in the copyright header at the top.

PR #817, opened 2023-12-23
fix: check if the secondary display active when only one is present

austinbhale:
Continuing off #816, the view count is 1 when it both is & isn't recording. I stepped through the code and found that it doesn't check if only one secondary display is active. For each frame, when it's inactive, we don't try to locate the device. When it's active, the swapchain gets updated! I can confirm this works for both active/inactive scenarios. Such a small change for the copyright header but I put that in there too :D


maluoi:
Hah! Okay, yes this looks _much_ more like the correct fix to me :D Thanks for digging further on that one!

PR #819, opened 2023-12-25
Get MSFT spatial anchor from SK virtual anchor

austinbhale:
For anchors created in the SK's "virtual" world, it would be helpful to have the "real" spatial anchor since certain UWP actions require it when moving between coordinate systems.

This can be used like the following:
```
bool result = anchor.TryGetPerceptionAnchor(out Windows.Perception.Spatial.SpatialAnchor spatialAnchor);
```

### DIFF
--- a/StereoKit/Assets/Anchor.cs
+++ b/StereoKit/Assets/Anchor.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 
 namespace StereoKit {
 	public class Anchor : IAsset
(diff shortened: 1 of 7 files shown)


StereoKit/Assets/Anchor.cs
```diff
@@ -15,7 +16,17 @@ public string Id
 		public BtnState Tracked => NativeAPI.anchor_get_tracked(_inst);
 		public bool Persistent => NativeAPI.anchor_get_persistent(_inst);
 		public string Name => NativeHelper.FromUtf8(NativeAPI.anchor_get_name(_inst));
-
+		public bool TryGetPerceptionAnchor<T>(out T spatialAnchor) where T : class
```
maluoi, line 19: I imagine this is expecting a pretty specific Type here? Thanks for using generics to avoid the library coupling! Could we also get commentdocs that indicate specifically what Type(s) should work here?

maluoi:
I think other than that one bit of info, this looks really good to me! If you can add that type info, I'll merge it in :)

StereoKit/Assets/Anchor.cs
austinbhale: Added! I think we're limited to the "class" constraint without a reference to winmd. I also figured to specify an instance of the SpatialAnchor type instead of its `ISpatialAnchor, ISpatialAnchor2` implementation since they're internal

PR #821, opened 2023-12-28
Theme refresh

maluoi:
## New Theme

This theme strives to maintain consistency with existing apps, while adding a bit of additional polish to clean up the look and feel of StereoKit apps!

- White exterior borders on the UI elements are removed, which helps reduce aliasing on lower resolution displays.
- Drop shadows are now on most UI elements, adding some pleasant gradients and grounding details.
- Text inputs now also have a new look, and are visually distinct from buttons. 

image

The previous proximity ring effect has always had issues with rounded surfaces, and I've never been exactly happy with how it looks. This iteration of the shader goes with a basic highlight glow for the finger. Simpler, but still effective, and a lot less visually noisy.

image

For those looking for a more adventurous departure from SK's current theme, try out a few adjustments to the layout settings! This snippet of code sets `rounding` to half the line height, making for a _very_ delightfully round UI!

```csharp
UI.Settings = new UISettings
{
	margin   = 0.009f,
	padding  = 0.012f,
	gutter   = 0.006f,
	depth    = 0.010f,
	rounding = 0.017f
};
```

image

## Previous Theme

The layout dimensions of the theme are _not_ changing with this refresh, so you should not expect to see any changes to the layout of your own UI! However, if you need to maintain the look of your app, you can find a recreation of the previous theme in this Gist. This may also be of interest as reference to those wishing to make deeper modifications to the look and feel of the UI.

https://gist.github.com/maluoi/d530747d50318efca7491f10df66d148

## API Additions

This adds `UI.GenQuadrantMesh`, which is a nice flexible mesh generator for [quadrantified button meshes](https://playdeck.net/blog/quadrant-sizing-efficient-ui-rendering). This allows you to define a button by specifying a "lathe" outline.

This does also change StereoKit's core UI shaders, removing the proximity circle, as well as simplifying some tinting code.

### DIFF
--- a/StereoKit/Native/NativeAPI.cs
+++ b/StereoKit/Native/NativeAPI.cs
@@ -709,6 +709,7 @@ namespace StereoKit
 
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   ui_quadrant_size_verts  ([In, Out] Vertex[] ref_vertices, int vertex_count, float overflow);
 		[DllImport(dll, CharSet = cSet, CallingConvention = call)] public static extern void   ui_quadrant_size_mesh   (IntPtr ref_mesh, float overflow);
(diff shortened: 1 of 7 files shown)


moshimeow:
Random note - I just updated to this and ended up needing to do a full rebuild of my project to un-cache the old shaders

PR #823, opened 2023-12-28
fix: check if the newly created MSFT anchor is tracked & overwrite its pose in the space

austinbhale:
Hi Nick, I'd like to check if an anchor is tracked immediately after its creation. One thing I'm not sure of is if you'd like the pose to be overwritten, even though it's very likely it'd output the same pose in that space?


maluoi:
I think this is fine! The pose gets overwritten on Step this way anyhow, so it'll keep the initial value consistent with what Step will assign it to :)

Issue #750, opened 2023-10-18
SK is blank after the first frame when running MRC on HL2


austinbhale:
## Description
Hey Nick, I found a pretty niche issue when launching an SK app while streaming MRC on the HL2. Reporting here since this would occur when customers want to stream the application for other people to see.

To reproduce,
1) Start Mixed Reality Capture from the Device Portal with `Render from Camera` enabled
2) Launch StereoKit app (latest dev)

Important Notes:
* MRC and `Render from Camera` works with an already running SK app
* The SK app renders for the first frame, then everything disappears and SK continues to step without rendering anything
* I couldn't reproduce this behavior with MSFT's [BasicXrApp](https://github.com/microsoft/OpenXR-MixedReality/blob/main/samples/BasicXrApp/OpenXrProgram.cpp), which leads me to believe this might be SK-related

## Platform / Environment
- UWP HoloLens 2
- Latest develop

## Logs or exception details
The initialization is successful and there is nothing notable to report from the logs. 
```
[SK diagnostic] Creating view: PrimaryStereo color:bgra32_sRGB depth:depth16 blend:Additive
[SK diagnostic] Set view: PrimaryStereo to 1440x936@1msaa
[SK diagnostic] Creating view: SecondaryMonoFPO_Msft color:bgra32_sRGB depth:depth16 blend:Alpha Blend
[SK diagnostic] Created unbounded app space at an offset of (0.0025,0.00017,-0.00034) (0,-0.0089,0,1)
[SK info] Initialization successful
```

After the first step, the following error repeats for [xrAcquireSwapchainImage](https://github.com/StereoKit/StereoKit/blob/6549a6cd9fb8a9168e7d70d1e98b0014601d0815/StereoKitC/xr_backends/openxr_view.cpp#L801) (throws [here](https://github.com/KhronosGroup/OpenXR-SDK/blob/55224479ab13db8ebc8ab1e3d49197bce6201b0b/src/loader/xr_generated_loader.cpp#L351)):
```
ERROR (OpenXR) : Invalid XrSwapchain handle with value 'XR_NULL_HANDLE'.
Exception thrown at 0x76BF56FF in SKApp.UWP.exe: Microsoft C++ exception: msxr::XrResultException at memory location 0x0470E6B0.
ERROR (OpenXR) : Invalid XrSwapchain handle with value 'XR_NULL_HANDLE'.
[...]

PR #830, opened 2024-01-02
Fix for #750 for the MSFT secondary view configuration

austinbhale:
This PR addresses 2 issues from #750:
1. If the camera is streaming/recording before the SK app initializes, the secondary view config is active. Since we assume [it starts as inactive](https://github.com/StereoKit/StereoKit/blob/092d717609425b8f9cb0bc4be00a3c53c91f2b32/StereoKitC/xr_backends/openxr_view.cpp#L310), the swapchain hadn't been updated
2. If the camera is active, there was an Access Violation error for invalid projection layer data. The data in the array should stay alive for the entire frame so it's not kept local

I can confirm StereoKitTest_UWP works for the following scenarios:
- HL2 is streaming before SK launches
- HL2 can stream on/off during an SK app session

Happy to make any other changes, thanks!

### DIFF
--- a/StereoKitC/xr_backends/openxr_view.cpp
+++ b/StereoKitC/xr_backends/openxr_view.cpp
@@ -159,7 +159,6 @@ void xr_extension_structs_free() {
 	xr_compositor_layers    .free();
 	xr_compositor_layer_sort.free();
 	xr_compositor_layer_ptrs.free();
-	xr_compositor_2nd_layer_ptrs.free();
 
 	xr_end_frame_chain_bytes .free();
 	xr_end_frame_chain_offset.free();
@@ -307,6 +306,7 @@ bool openxr_views_create() {
 			device_display_t display = {};
 			if (!openxr_display_create(types[t], &display))
 				return false;
+			display.active = false;
 
 			XrSecondaryViewConfigurationStateMSFT state = { XR_TYPE_SECONDARY_VIEW_CONFIGURATION_STATE_MSFT };
 			state.active                = false;
@@ -369,6 +369,7 @@ void openxr_views_destroy() {
 	xr_displays_2nd      .free();
 	xr_display_2nd_states.free();
 	xr_display_2nd_layers.free();
+	xr_compositor_2nd_layer_ptrs.free();
 	xr_extension_structs_free();
 }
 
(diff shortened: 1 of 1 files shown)


StereoKitC/xr_backends/openxr_view.cpp
```diff
@@ -158,6 +159,7 @@ void xr_extension_structs_free() {
 	xr_compositor_layers    .free();
 	xr_compositor_layer_sort.free();
 	xr_compositor_layer_ptrs.free();
+	xr_compositor_2nd_layer_ptrs.free();
```
maluoi, line 162: This probably belongs in `openxr_views_destroy` instead!

StereoKitC/xr_backends/openxr_view.cpp
```diff
@@ -704,12 +706,10 @@ bool openxr_render_frame() {
...
 	for (int32_t i = 0; i < xr_displays_2nd.count; i++) {
-		if (xr_displays_2nd[i].active != (bool32_t)xr_display_2nd_states[i].active) {
-			xr_displays_2nd[i].active  = (bool32_t)xr_display_2nd_states[i].active;
+		xr_displays_2nd[i].active  = (bool32_t)xr_display_2nd_states[i].active;
 
-			if (xr_displays_2nd[i].active) {
-				openxr_display_swapchain_update(&xr_displays_2nd[i]);
-			}
+		if (xr_displays_2nd[i].active) {
+			openxr_display_swapchain_update(&xr_displays_2nd[i]);
```
maluoi, line 712: Was this chunk a critical part of the fix? Since `openxr_display_swapchain_update` does also have some checks to prevent redundant updates, this is probably okay. However, [the spec](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#_secondary_view_configuration_properties) does kinda indicate that the `xrEnumerateViewConfigurationViews` call in `openxr_display_swapchain_update` should only really be necessary when `active` changes.

austinbhale: This is only critical if the secondary display is active prior to SK running. If we set the display as initially inactive (same as `XrSecondaryViewConfigurationStateMSFT`), then we can keep this check! That works too :)

maluoi:
Awesome, ty! Glad this is back in working order again :)

PR #834, opened 2024-01-11
Add list mode for File picker and sorting by file size and add sprite…

ponugotimanasaqc:
… for folders.


StereoKitC/tools/file_picker.cpp
```diff
@@ -53,13 +53,16 @@ namespace sk {
...
 	bool  file;
+	long size;
 };
 
 struct fp_path_t {
 	char          *folder;
 	array_t<char*> fragments;
 };
 
+enum FpSortBy { Name, Size };
```
maluoi, line 64: Naming convention, should be `fp_sort_by_`, etc.

StereoKitC/tools/file_picker.cpp
```diff
@@ -53,13 +53,16 @@ namespace sk {
 struct fp_item_t {
 	char *name;
 	bool  file;
+	long size;
```
maluoi, line 56: I would prefer using some variant of `int64_t` or `int32_t` instead of `long` or `int` (throughout this PR)! `long` and `int` have some amount of ambiguity in C/C++.

StereoKitC/tools/file_picker.cpp
```diff
@@ -75,6 +78,10 @@ char                         fp_buffer[1023];
 bool                         fp_call          = false;
 void                        *fp_call_data     = nullptr;
 bool                         fp_call_status   = false;
+bool                         fp_files_sorted_asc = true;
+FpSortBy                     fp_sortby = Name;
+bool                         sort_order_changed = false;
```
maluoi, line 83: Prefix with `fp_` for consistency

StereoKitC/tools/file_picker.cpp
```diff
@@ -435,30 +461,114 @@ void file_picker_update() {
...
+				if (ui_button_sz("Name", size)) {
+					fp_sortby = Name;
+					sort_order_changed = true;
+				}
+			}
+			ui_sameline();
+			ui_text_sz(" ", text_align_x_center, text_fit_clip, size);
+			ui_sameline();
+			if (fp_sortby == Size) {
+				if (ui_button_img_sz("Size", spr_up, layout, size)) {
```
maluoi, line 500: Was there a path to do this using ui_toggle with images? Or was something blocking that?

StereoKitC/tools/file_picker.cpp
```diff
@@ -435,30 +461,114 @@ void file_picker_update() {
...
+				}
+			}
+			else {
+				if (ui_button_sz("Name", size)) {
+					fp_sortby = Name;
+					sort_order_changed = true;
+				}
+			}
+			ui_sameline();
+			ui_text_sz(" ", text_align_x_center, text_fit_clip, size);
```
maluoi, line 497: Spacing like this can vary based on the font used! I'd prefer to see spacing done by calculating more specific sizes for elements, and obey the theme's rules for gutters. See `ui_layout_remaining` and `ui_get_settings`.

StereoKitC/tools/file_picker.cpp
```diff
@@ -435,30 +461,114 @@ void file_picker_update() {
...
+			ui_sameline();
+			if (fp_active == nullptr) ui_push_enabled(false);
 			ui_label(fp_active?fp_active:"None selected...");
 			if (fp_active == nullptr) ui_pop_enabled();
 		} break;
 		}
 
 		ui_hseparator();
 
+		sprite_t spr_folder = sprite_find(ui_default_id_spr_arrow_right);
```
maluoi, line 477: This should be done once and cached rather than referenced every step. This is also a reference counted asset that should be released when finished!

StereoKitC/tools/file_picker.cpp
ponugotimanasaqc: ui toggle can work but ui wouldnt look great. current image depicts the sort in ascending order. In future if needed, we can add sort by descending as well

StereoKitC/tools/file_picker.cpp
maluoi: In this case, consistent convention for enums would be:

```cpp
enum fp_sort_by_ {
    fp_sort_by_name,
    fp_sort_by_size
};
```

StereoKitC/tools/file_picker.cpp
maluoi: File sizes seem to cap out around 16TB, which I believe don't fit in just a 32 bit integer, so I believe this specifically should probably still be an `int64_t`

StereoKitC/tools/file_picker.cpp
```diff
@@ -435,30 +467,118 @@ void file_picker_update() {
...
-			} else if (ui_button_sz(fp_items[i].name, size)) {
-				if (fp_items[i].file)
-					fp_active = fp_items[i].name;
+			if (fp_list_mode) {
+				if (i >= fp_items.count) {
+					ui_layout_reserve(size, false, 0.0f);
+					ui_sameline();
+					ui_layout_reserve(size, false, 0.0f);
+					ui_sameline();
+					ui_layout_reserve(size, false, 0.0f);
```
maluoi, line 545: For cleanliness sake, I'd prefer just one call to ui_layout_remaining with the correct size, instead of 3.

ponugotimanasaqc: Tried and tested with ui_layout_remaining. It is not able to correctly place the scroll down button. The original Non-list mode assumes 5 rows of data and places the scroll down button. To keep in parity with the same design choice in list mode, I am adding empty space to accommodate all the rows so that, the scroll down button is in sync with the panel

StereoKitC/platforms/android.cpp
```diff
@@ -495,7 +495,7 @@ font_t platform_default_font() {
 
 ///////////////////////////////////////////
 
-void platform_iterate_dir(const char* directory_path, void* callback_data, void (*on_item)(void* callback_data, const char* name, bool file)) {}
+void platform_iterate_dir(const char* directory_path, void* callback_data, void (*on_item)(void* callback_data, const char* name, bool file, const int32_t size)) {}
```
maluoi, line 498: Since we're using `int64_t` for file size, it should be consistent throughout to avoid data loss.

maluoi: Also, thinking at a higher level, is it appropriate to be passing file size through `platform_iterate_dir`? If we expand the file picker to say... include file modification and creation date, it might not make sense to be providing this data via the file discovery callback. Might be better instead to look it up in the file picker?

StereoKitC/tools/file_picker.cpp
```diff
@@ -427,6 +459,10 @@ void file_picker_update() {
 				fp_call_status = true;
 			}
 			ui_sameline();
+			if (ui_button("ListMode")) {
```
maluoi, line 462: I'm not sure I really liked the way this looked in the UI, can you play around with the location some, and see what else can be done for it? This should also probably be some kind of toggle rather than button.

StereoKitC/tools/file_picker.cpp
```diff
@@ -435,30 +471,114 @@ void file_picker_update() {
...
+			if (fp_sort_order_changed) {
+				switch (fp_sortby) {
+				case fp_sort_by_name:
+					fp_items.sort([](const fp_item_t& a, const fp_item_t& b) { return a.file != b.file ? a.file - b.file : strcmp(a.name, b.name); });
+					break;
+				case fp_sort_by_size:
+					fp_items.sort([](const fp_item_t& a, const fp_item_t& b) { return a.size == b.size ? 0 : a.size - b.size > 0 ? 1 : -1;});
+					break;
+				}
+			}
```
maluoi, line 535: Could we also get ascending/descending sort?

maluoi:
image

maluoi:
UIEdits

- I liked how name/size buttons lined up with the width of their columns previously, let's keep that behavior.
- Can we line up the ListMode button with the Up/Down arrows as well? Maybe change it to just "List". It's okay if it's squished, I think eventually I'll want to make an icon or two for this anyhow.
- In List mode, there seems to be a new gap to the right of the back panel, between it and the Up/Down arrows.
- When selecting a folder, sorting options change. I think I'd prefer it to stay the same throughout.
- File size should alight right, can we maybe format it as kilobytes instead of bytes?
- Can we use the Up/Down sprites for the Up/Down buttons as well?
- For folders, maybe lets also use the `sk/ui/toggle_off` sprite instead of the right arrow.

PR #840, opened 2024-01-17
WIP: Magic Leap 2 Integration

austinbhale:
Hey Nick, the Magic Leap team is loaning us a device so I'll be working on its integration in my spare time. I'll be keeping this PR updated in the meantime, so that it's easier to show you updates and ask questions.

## Progress
* [StereoKit's first run on Magic Leap 2](https://youtu.be/35YIcJs6i0s)

## Supported Extensions
🔴 not supporting in this PR
🟡 work in progress
🟢 complete
❔ unsure if StereoKit does/will support it?

| Extension | Supported |
| ------------- | ------------- |
| XR_EXT_eye_gaze_interaction | 🟡 |
| XR_EXT_hand_interaction    | ✅  |
| XR_EXT_hand_tracking      | ✅  |
| XR_EXT_local_floor        | ✅  |
| XR_EXT_palm_pose          | ✅  |
| XR_KHR_android_create_instance | ✅ |
| XR_KHR_composition_layer_depth | ✅ |
| XR_KHR_convert_timespec_time | ✅ |
| XR_KHR_opengl_es_enable   | ✅  |
| XR_KHR_vulkan_enable      | ❔  |
| XR_KHR_vulkan_enable2     | ❔  |
| XR_ML_compat              | 🔴  |
| XR_ML_frame_end_info      | 🔴  |
| XR_ML_global_dimmer       | 🟡  |
| XR_ML_localization_map    | ❔  |
| XR_ML_marker_understanding| ❔  |
| XR_ML_ml2_controller_interaction | 🟡 |
| XR_ML_user_calibration    | 🔴  |
| XR_MND_headless           | ❔  |
| XR_MSFT_unbounded_reference_space | ✅ |
| XR_EXT_view_configuration_depth_range | ❔ |

## Known issues
* Very apparent jitter (& sometimes colorful artifacts) when you either move your head too quickly or move around.
This may be a limitation of the on-device SLAM, which causes perceived motion of stationary content. For example, reading SK text while moving around becomes illegible as it seemingly "repeats" multiple times until the tracking restabilizes. This is probably why ML2 offers advice on [how to mitigate user micro-movements](https://developer-docs.magicleap.cloud/docs/guides/best-practices/comfort-content-placement/#how-to-mitigate-user-micro-movements) for UI. If StereoKit is in immediate mode, is it correct to assume there's no buffer in place?
*Please note that this visual artifact won't be seen in the mixed reality video.*
* Eyes demo not working
* Only Release mode works for `StereoKitTest_NetAndroid`

## Initialization Log
```
[StereoKit] Initializing StereoKit v0.3.9 Android x64...
[StereoKit] Initializing Platform
[StereoKit] [sk_gpu] EGL version 1.5
[StereoKit] [sk_gpu] Using OpenGL: OpenGL ES 3.2 Mesa 20.1.2 (git-6582e67fb7)
[StereoKit] [sk_gpu] Device: AMD MERO (DRM 3.37.0, 5.4.19-android-x86_64, LLVM 10.0)
[StereoKit] [sk_gpu] Debug info enabled.
[StereoKit] Starting a XR mode app
[StereoKit] Runtime OpenXR Extensions:
[StereoKit] ___________________________________
[StereoKit] |     Usage | Extension
[StereoKit] |-----------|----------------------
[StereoKit] |   present | XR_EXT_conformance_automation
[StereoKit] |   present | XR_EXT_debug_utils
[StereoKit] |   present | XR_EXT_plane_detection
[StereoKit] |   present | XR_EXT_view_configuration_depth_range
[StereoKit] |   present | XR_KHR_opengl_enable
[StereoKit] |   present | XR_KHR_vulkan_enable
[StereoKit] |   present | XR_KHR_vulkan_enable2
[StereoKit] |   present | XR_ML_compat
[StereoKit] |   present | XR_ML_frame_end_info
[StereoKit] |   present | XR_ML_global_dimmer
[StereoKit] |   present | XR_ML_marker_understanding
[StereoKit] |   present | XR_ML_ml2_controller_interaction
[StereoKit] |   present | XR_ML_private
[StereoKit] |   present | XR_ML_session_performance_info
[StereoKit] |   present | XR_ML_system_notifications
[StereoKit] |   present | XR_ML_user_calibration
[StereoKit] |   present | XR_MND_headless
[StereoKit] | ACTIVATED | XR_EXT_eye_gaze_interaction
[StereoKit] | ACTIVATED | XR_EXT_hand_interaction
[StereoKit] | ACTIVATED | XR_EXT_hand_tracking
[StereoKit] | ACTIVATED | XR_EXT_local_floor
[StereoKit] | ACTIVATED | XR_EXT_palm_pose
[StereoKit] | ACTIVATED | XR_KHR_android_create_instance
[StereoKit] | ACTIVATED | XR_KHR_composition_layer_depth
[StereoKit] | ACTIVATED | XR_KHR_convert_timespec_time
[StereoKit] | ACTIVATED | XR_KHR_opengl_es_enable
[StereoKit] | ACTIVATED | XR_MSFT_unbounded_reference_space
[StereoKit] |___________|______________________
```

### DIFF
--- a/Examples/StereoKitTest/StereoKitTest_NetAndroid/AndroidManifest.xml
+++ b/Examples/StereoKitTest/StereoKitTest_NetAndroid/AndroidManifest.xml
@@ -54,4 +54,11 @@
 	<uses-feature    android:name="oculus.software.face_tracking"     android:required="false"/>
 	<uses-feature    android:name="oculus.software.eye_tracking"      android:required="false"/>
 	<uses-feature    android:name="com.oculus.experimental.enabled"   android:required="false"/>
+
+	<!-- Magic Leap 2 specific items -->
+	<uses-permission android:name="com.magicleap.permission.EYE_TRACKING" />
+	<uses-permission android:name="com.magicleap.permission.HAND_TRACKING" />
+	<uses-permission android:name="com.magicleap.permission.SPATIAL_MAPPING" />
+	<uses-feature    android:name="com.magicleap.api_level" android:version="21" />
+
 </manifest>
--- a/Examples/StereoKitTest/StereoKitTest_NetAndroid/StereoKitTest_NetAndroid.csproj
+++ b/Examples/StereoKitTest/StereoKitTest_NetAndroid/StereoKitTest_NetAndroid.csproj
@@ -16,8 +16,8 @@
 		<SKAssetDestination>Assets</SKAssetDestination>
 		<SKShowDebugVars>true</SKShowDebugVars>
 		
-		<!--<TestBuildMode>x64</TestBuildMode>-->
-		<TestBuildMode>ARM64</TestBuildMode>
+		<TestBuildMode>x64</TestBuildMode>
+		<!--<TestBuildMode>ARM64</TestBuildMode>-->
 
 		<RuntimeIdentifiers Condition="'$(TestBuildMode)'=='ARM64'">android-arm64</RuntimeIdentifiers>
 		<RuntimeIdentifiers Condition="'$(TestBuildMode)'=='x64'">android-x64</RuntimeIdentifiers>


maluoi:
Hey! Nice to see this pop up, that looks pretty decent in the video! :) I believe I've seen _someone_ running SK on ML2 before, but only briefly, this is a really nice run-down.

Do you have any info/logs on why Debug mode isn't working?

Their comments on "buffers" are phrased really weird. I think it's unrelated to any stuttering, I believe they're just trying to say that UI shouldn't move on its own when the user is trying to interact with it? Which is fine, StereoKit's core stuff doesn't move anything unless directed to by the user. The stuttering does sound concerning though, does it happen throughout the device, or is it specific to SK apps?

If eye gaze isn't working, is it possible the device needs to [explicitly request permission](https://github.com/StereoKit/StereoKit/blob/master/Examples/StereoKitTest/StereoKitTest_Xamarin/MainActivity.cs#L61-L64) with a pop-up? This is how eye tracking currently works on Quest Pro.

Does StereoKit already recognize the ML2 controllers at all? In theory if their input system is well built, the runtime should map the controllers to one of SK's existing bindings. You would be able to see a controller rendered in the Controller demo, and it would log the name of the input profile it uses.

Also _very_ nice to see ML already using `XR_EXT_plane_detection`! I should look into supporting that for sure!

paulmelis:
> Very apparent jitter (& sometimes colorful artifacts) when you either move your head too quickly or move around.
This may be a limitation of the on-device SLAM, which causes perceived motion of stationary content. For example, reading UI text while moving around becomes illegible as it seemingly "repeats" multiple times until the tracking restabilizes.

Oh thank god somebody is seeing the same issues I am! We've had our ML2 replaced *twice* by Magic Leap because of these issues and I could find nobody else with an ML2 who was experiencing this kind of ghosting.

austinbhale:
> Do you have any info/logs on why Debug mode isn't working?

Just updated the failed log for Debug mode! I've also tried disabling "Fast Deployment" with no luck

> This is how eye tracking currently works on Quest Pro

You're right, it's the same here too! The eye tracking & spatial mapping permissions are considered dangerous, so they must be granted at runtime: [App Permissions](https://www.magicleap.care/hc/en-us/articles/6653784345741-App-Permissions). I put in a fix for granting those on the main thread once SK is running

> The stuttering does sound concerning though, does it happen throughout the device, or is it specific to SK apps?

Turns out it's not specific to SK apps! Though it's more noticeable than their samples (C API & OpenXR). Even in the main menu, they seem to account for the effect by lerping the menu's orientation & translating a bit delayed so it appears "smoother" (video below). So I don't think there is much we can do here except put in a request to ML about it

ml2-menu-movement

> Does StereoKit already recognize the ML2 controllers at all?

I'm not seeing it from the log 😬 I'll look into trying to enable it, but Debug mode would be helpful here

> Also very nice to see ML already using XR_EXT_plane_detection!

Support for [spatial perception](https://developer-docs.magicleap.cloud/docs/guides/unreal/unreal-supported-features/#perception) is still experimental but their sample for `XR_EXT_plane_detection` looked good!

I updated the list for what this PR could contain. I guess only controller support is needed at the moment? I thought about also supporting the occlusion filter and spatial anchors but it seems too experimental in their OpenXR code. Global dimmer would be fun to implement into SK too if you'd want it

maluoi:
Thanks for the logs! These two lines stand out:
```
[linker] library "/system/lib64/libopenxr_layer_core_validation.so" ("/system/lib64/libopenxr_layer_core_validation.so") needed or dlopened by "/data/app/net.stereokit.stereokittest_netandroid-ATnHt-yg11pKXGgJRbFHNQ==/lib/x86_64/libopenxr_loader.so" is not accessible for the namespace: [name="classloader-namespace", ld_library_paths="", default_library_paths="/data/app/net.stereokit.stereokittest_netandroid-ATnHt-yg11pKXGgJRbFHNQ==/lib/x86_64:/data/app/net.stereokit.stereokittest_netandroid-ATnHt-yg11pKXGgJRbFHNQ==/base.apk!/lib/x86_64", permitted_paths="/data:/mnt/expand:/data/data/net.stereokit.stereokittest_netandroid"]
[StereoKit] Couldn't create OpenXR instance [XR_ERROR_FILE_ACCESS_ERROR], is OpenXR installed and set as the active runtime?
```

It looks like the OpenXR Loader is trying to load `libopenxr_layer_core_validation.so`, a system library, and isn't getting permissions for it? This feels kinda reminiscent of a problem I ran into with Oculus a while back.

Not sure if it's the same sort of issue, or if this is even the correct fix for it, but it might be worth trying. I added [this line to the manifest](https://github.com/StereoKit/StereoKit/blob/master/Examples/StereoKitTest/StereoKitTest_NetAndroid/AndroidManifest.xml#L9), I'd try the same thing, but with `libopenxr_layer_core_validation.so`? Maybe?

> I'm not seeing it [controllers] from the log 😬 I'll look into trying to enable it, but Debug mode would be helpful here

There's also the in-app log in StereoKitTest's hand menu :) But yeah, if you need to add the profile, check [something like this one for reference](https://github.com/StereoKit/StereoKit/blob/master/StereoKitC/xr_backends/openxr_input.cpp#L343-L386). Someday I should make those customizable from top-level, but it's not too tricky to add there.

> I thought about also supporting the occlusion filter and spatial anchors but it seems too experimental in their OpenXR code. Global dimmer would be fun to implement into SK too if you'd want it

Which were the occlusion filter and spatial anchors? I didn't quite spot those. I see `XR_ML_marker_understanding`, which is... April Tags. Not my favorite markers, but I guess they're practical.

Global dimmer is pretty intriguing! I have no idea where that would go though :D I'm tempted to say it might be better implemented at top level, but if you can find a good spot for it in the core, I don't think I'd object.

austinbhale:
BTW, just captured a better depiction of the quite nauseating effect from ML2. It's difficult to see on camera but this is pretty much the effect that your eyes see! It also happens in the sample apps, especially when reading GUI text

image

https://github.com/StereoKit/StereoKit/assets/32556981/30f0c50e-1320-431e-9a53-d76d4d918a75

maluoi:
This reminds me a bit of the HoloLens 2 scanning artifacts! I wonder if maybe it's a similar sort of thing? IIRC, the R, G, and B channels of the HL2 display are actually drawn separately at different times! So you could see colored ghosts if you were moving your head quickly.

ML2 is similarly using a waveguide display, but I don't really know the details of their implementation. It's possible this might be a similar issue with a more artifact-ey implementation, or it could be off in a different direction and be some display persistence thing? I'm not really sure!

paulmelis:
> This reminds me a bit of the HoloLens 2 scanning artifacts! I wonder if maybe it's a similar sort of thing? IIRC, the R, G, and B channels of the HL2 display are actually drawn separately at different times! So you could see colored ghosts if you were moving your head quickly.

You can indeed easily see the color separation with head movement, and GUI text as Austin mentioned. But the weird thing I've found is that the ghosting doesn't happen in all situations (and it doesn't appear to be frame-rate related). For example, just loading the google home page in the browser app easily shows color separation in the google logo, with nothing particularly graphics-heavy loaded. But take another graphics-light situation, such as the LeapBrush app, which allows you to draw lines/ribbons in 3D using the controller. When you draw a few lines and move your head around these lines there's hardly any ghosting in them. But the app also show help text on the controller buttons and moving the controller in your field-of-view, while keeping your head still, very much shows ghosting of that text. Yet, put the controller stationary in a place (e.g. on a table) and make the same relative movements with your head and no ghosting is visible on the text.

austinbhale:
I talked to some representatives from Magic Leap 2 and this is what they mentioned:
- Since these are world-locked holograms, the device tries to predict a user’s movement and has some interpolation error when the movement is unexpected
- User should specify a focus point on a stabilization plane (e.g., via eye tracking)
- User should specify the near and far clipping planes
- The R, G, B channels of the display are drawn at different times
- The refresh rate is upscaled from 90 -> 120 hz, and the pixels don’t turn off fast enough

The Graphics team is constantly trying to improve this effect, with the goal that the developer doesn't have to manually set a focus distance by default.

For this PR, I'm waiting on their recommended spatial anchor and controller interaction OpenXR samples.

Another tidbit I learned is about their future release of aligning holograms to the camera image. For their version of mixed reality capture, they'll be having a 3rd-party plugin in Unity for this alignment. It always sounds like some extra hurdles for them to fully focus on OpenXR, since their customers are primarily using Unity. However, they are sharing StereoKit around internally and even had some more requests for it! :)

paulmelis:
@austinbhale Nice feedback! Some questions though:

> Since these are world-locked holograms, the device tries to predict a user’s movement and has some interpolation error when the movement is unexpected

I guess this is the main reason I see the different color separation behaviour with LeapBrush and the controller overlay, as described above. Even when using very erratic head movement and the controller lying stationary I see no color separation. If it really was in the display tech itself it should be noticeable even in this situation.

Makes me wonder if they do position prediction during the RGB cycle, but it simply is not precise enough, hence the color separation.

> User should specify a focus point on a stabilization plane (e.g., via eye tracking)

Is this specific to the Unity integration for ML2? I mean, how would you even do this in SK?

> The refresh rate is upscaled from 90 -> 120 hz, and the pixels don’t turn off fast enough

Have you managed to get above 60 FPS rendering using StereoKit? So far, I only manage 60FPS, even with simple scenes (based on a simple custom FPS counter).

maluoi:
> Since these are world-locked holograms, the device tries to predict a user’s movement and has some interpolation error when the movement is unexpected

On HoloLens, we called this Late Stage Reprojection, and there were two methods of doing it. One was to submit a depth buffer to the compositor, and let the compositor reproject based on the depth at each pixel. The other (older) one was to just pick a point, or plane, and let the compositor reproject as if everything was on that plane.

When we did have to specify a plane, it was a UWP API call, I don't believe the point/plane method survived to make it into OpenXR. The depth method seems to have pretty much superseded the plane/point method by this point in time! I guess I'd probably be looking for an ML API for this, or an unpublished OpenXR extension?

> Another tidbit I learned is about their future release of aligning holograms to the camera image. For their version of mixed reality capture, they'll be having a 3rd-party plugin in Unity for this alignment.

HoloLens did this one through `XR_MSFT_first_person_observer`, which both Unity and SK already support :( I wish it had a better name.

> Have you managed to get above 60 FPS rendering using StereoKit? So far, I only manage 60FPS, even with simple scenes (based on a simple custom FPS counter).

If you're locked at 60, I might hunt around for some ML specific Android Manifest items?

maluoi:
Ohhh, I see. Render at 60hz, reproject to 120hz. Might even be a side-effect or "number magic" related to the way they draw R G and B channels separately.

austinbhale:
Closing this since I don't have a device 😎

PR #846, opened 2024-01-31
fix: Always build StereoKit's C# library as AnyCPU

austinbhale:
This allows to build a project targeting a different architecture without having to rebuild StereoKit for that architecture
image
This changes the output path from `bin\netstandard2.0` to `bin\x64\Debug\netstandard2.0`


maluoi:
Okay, this is interesting, and I think I see a bit more about what's going on here! In theory, a netstandard2.0 project is supposed to be framework _and_ architecture independent, so the fact that this needs to be built for different architectures is a bit surprising to me. I _have_ seen some warnings that might have been able to cue me in on this, but I had no idea what was going on there at the time, and nothing ever seemed broken.

Soooo, I took a peek in a C# disassembler, aaaand 🙀 
image

Looks like it does actually have an x64 tag in the metadata! Alright, today's one of those days where I'm not too sure why it was ever okay with that, this is just how StereoKit _has_ been shipping all this time, for all platforms/archs. It looks like `msil` should be the appropriate value here?

With a bit of coaxing and configuration, I _can_ get it to output an `msil` tagged dll instead. I still need to test this a bit further and make sure it doesn't break other stuff, but this feels way more right.

image

Would that work for you, or is there some other thing that gets in the way here?

maluoi:
It does look like public NuGet builds of SK are properly shipping as msil DLLs, so this may be something more specifically about how local builds are done...

austinbhale:
Woa no wonder! After your automated builds, this probably isn't a common use case. Even though it takes a few seconds to rebuild, it feels like such a relief to switch archs so seamlessly 😄 I thought Any CPU would've taken care of it but that caused config mapping errors in Visual Studio. How can I test your changes -- set the PlatformTarget to msil, or is there a branch?

austinbhale:
Setting the `PlatformTarget` to `AnyCPU` seems to work for all projects over here after clearing the cache in "postSolution" and reopening Visual Studio! (suggested [here](https://stackoverflow.com/a/74358346)) I had gotten the skipping too until doing this since reloading projects didn't perform this step

maluoi:
Heyyy, yeah that sounds promising! I had tried clearing the cache part of the .sln before, to mixed results, but I wasn't using AnyCpu (No spaces! Augh) at the time! I'll double check this stuff locally :)

maluoi:
Awesome 🥳 thanks for helping lay that mystery to rest!

PR #873, opened 2024-02-22
Initial commit for Keyboard customization API

ponugotimanasaqc:


StereoKitC/tools/virtual_keyboard.cpp
```diff
@@ -53,12 +53,14 @@ uint64_t                  keyboard_go_toggle        = 0;
 int32_t                   keyboard_visit_return_idx = -1;
 text_context_             keyboard_text_context     = text_context_text;
 
-array_t<keylayout_info_t> keyboard_layers           = {};
-int32_t                   keyboard_active_root      = 0;
-int32_t                   keyboard_active_idx       = 0;
-int32_t                   keyboard_ctx_root_text    = 0;
-int32_t                   keyboard_ctx_root_number  = 0;
-int32_t                   keyboard_ctx_root_uri     = 0;
+hashmap_t<text_context_, array_t<keylayout_info_t>> keyboards = {};
```
maluoi, line 56: Maybe we can shift `text_context_` to be just a consecutive 0,1,2,3.., and then use that to index a normal array of `array_t`? `hashmap_t` is okay, but reallllly feels like overkill here.

StereoKitC/tools/virtual_keyboard.cpp
```diff
@@ -198,12 +206,34 @@ spr:sk/ui/shift---3-go_0|*|=|+|#|%|'|"|spr:sk/ui/backspace-\b--3
...
 
 ///////////////////////////////////////////
 
+bool virtualkeyboard_switch(const char** keyboard_layouts, int layouts_num, text_context_ keyboard_type) {
+	keylayout_info_t layer = {};
+	array_t<keylayout_info_t> keyboard_layers = {};
+	for (int i = 0; i < layouts_num; i++) {
+		if (virtualkeyboard_parse_layout(*(keyboard_layouts + i), (int32_t)strlen(*(keyboard_layouts + i)), &layer)) {
+			keyboard_layers.add(layer);
+		}
```
maluoi, line 221: `else` this should return false here!

StereoKitC/tools/virtual_keyboard.cpp
```diff
@@ -144,39 +149,42 @@ Tab-\t-9-3|q-q-81|w-w-87|e-e-69|r-r-82|t-t-84|y-y-89|u-u-85|i-i-73|o-o-79|p-p-80
...
+	keyboard_layers = {};
 	////// URI keyboard //////
 	layout =
 R"(`-`-192|1-1-49|2-2-50|3-3-51|4-4-52|5-5-53|6-6-54|7-7-55|8-8-56|9-9-57|0-0-48|\--\--189|=-=-187|spr:sk/ui/backspace-\b-8-3|---2|spr:sk/ui/close----close
 Tab-\t-9-3|q-q-81|w-w-87|e-e-69|r-r-82|t-t-84|y-y-89|u-u-85|i-i-73|o-o-79|p-p-80|[-[-219|]-]-221|\\-\\-220|.com-.com--4
 Enter-\n-13-4|a-a-65|s-s-83|d-d-68|f-f-70|g-g-71|h-h-72|j-j-74|k-k-75|l-l-76|;-;-186|'-'-222|Enter-\n-13-3|.net-.net--4
 spr:sk/ui/shift--16-5-visit_1|z-z-90|x-x-88|c-c-67|v-v-86|b-b-66|n-n-78|m-m-77|,-,-188|.-.-190|/-/-191|spr:sk/ui/shift--16-2-visit_1|spr:sk/ui/arrow_up--38|https://-https://--4
 Ctrl--17-4-mod|Cmd--91-3|Alt--18-3-mod| - -32-9|Alt--18-3-mod|Ctrl--17-3-mod|spr:sk/ui/arrow_left--37|spr:sk/ui/arrow_down--40|spr:sk/ui/arrow_right--39|)";
-	if (virtualkeyboard_parse_layout(layout, (int32_t)strlen(layout), &layer)) keyboard_ctx_root_uri = keyboard_layers.add(layer);
+	if (virtualkeyboard_parse_layout(layout, (int32_t)strlen(layout), &layer)) keyboard_layers.add(layer);
```
maluoi, line 169: Could we switch this whole section to use the new public API for setting the keyboard layout? Then we know for sure that public API is getting some internal usage and testing too :)

StereoKit/Native/NativeEnums.cs
```diff
@@ -5,740 +5,766 @@
 using System;
```
maluoi, line 5: This file is actually generated automatically by parsing `stereokit.h`! If you run the StereoKitAPIGen project in the solution, you'll see this file update automatically :)

maluoi: Since this file wasn't changed, could you restore it to its initial state? Running StereoKitAPIGen  should do this quickly.

StereoKit/Util/Platform.cs
```diff
@@ -48,9 +48,23 @@ public static bool KeyboardVisible
...
+		/// and use a fallback keyboard when not. On systems with physical
+		/// keyboards, soft keyboards generally will not be shown if the user
+		/// has interacted with their physical keyboard recently.</summary>
+		/// <param name="show">Tells whether or not to open or close the soft
+		/// keyboard.</param>
+		/// <param name="inputType">Soft keyboards can change layout to
+		/// optimize for the type of text that's required. StereoKit will
+		/// request the soft keyboard layout that most closely represents the
+		/// TextContext provided.</param>
+		public static bool KeyboardSwitch(TextContext keyboardType, string[] keyboardLayout)
```
maluoi, line 64: There's a number of warnings related to commentdocs here.

StereoKitC/tools/virtual_keyboard.cpp
```diff
@@ -112,10 +110,10 @@ void virtualkeyboard_open(bool32_t open, text_context_ type) {
...
 	switch (type) {
-	case text_context_text:  keyboard_active_root = keyboard_ctx_root_text;   break;
-	case text_context_number:keyboard_active_root = keyboard_ctx_root_number; break;
-	case text_context_uri:   keyboard_active_root = keyboard_ctx_root_uri;    break;
-	default:                 keyboard_active_root = keyboard_ctx_root_text;   break;
+	case text_context_text:  keyboard_active_root = text_context_text;   break;
+	case text_context_number:keyboard_active_root = text_context_number; break;
+	case text_context_uri:   keyboard_active_root = text_context_uri;    break;
+	default:                 keyboard_active_root = text_context_text;   break;
 	}
```
maluoi, line 117: Does this still need to be a switch/case? This can probably be done much more easily with an `if` now.

StereoKitC/stereokit.h
```diff
@@ -1741,6 +1741,7 @@ SK_API bool32_t platform_keyboard_get_force_fallback(void);
 SK_API void     platform_keyboard_set_force_fallback(bool32_t force_fallback);
 SK_API void     platform_keyboard_show              (bool32_t visible, text_context_ type);
 SK_API bool32_t platform_keyboard_visible           (void);
+SK_API bool32_t platform_keyboard_switch(text_context_ type, char** keyboard_layout, int layouts_num);
```
maluoi, line 1744: I would still prefer to see this as `platform_keyboard_set_layout`/`Platform.KeyboardSetLayout`, unless you have some justification for why `switch` is a better name?

StereoKitC/tools/virtual_keyboard.cpp
```diff
@@ -130,53 +128,49 @@ bool virtualkeyboard_get_open() {
...
-	layout =
+	layouts[0] =
 R"(7|8|9|spr:sk/ui/backspace-\b
 4|5|6|\-
 1|2|3
 0|.|Return-\n--4-close|)";
-	if (virtualkeyboard_parse_layout(layout, (int32_t)strlen(layout), &layer)) keyboard_ctx_root_number = keyboard_layers.add(layer);
-
+	layouts[1] = nullptr;
+	virtualkeyboard_switch(text_context_number, layouts, 1);
```
maluoi, line 173: This is better, but could we use the public version of this function instead of the internal one?

StereoKitC/tools/virtual_keyboard.cpp
```diff
@@ -197,13 +191,38 @@ R"(1|2|3|4|5|6|7|8|9|0
...
+		}
+	}
+
+	if (keyboard_layers.count != 0) {
+		switch (keyboard_type) {
+			case text_context_text:  keyboards[0] = keyboard_layers;;   break;
+			case text_context_number: keyboards[1] = keyboard_layers;; break;
+			case text_context_uri:    keyboards[2] = keyboard_layers;;    break;
+			default:                 keyboards[0] = keyboard_layers;;   break;
+		}
```
maluoi, line 218: This too, does this still need to be a switch/case, or can it be an `if`?

maluoi: I'd also love to see a bit more attention to detail on spacing and unnecessary syntax. It's fine if it doesn't look exactly like my code, but it should at least be clean and consistent. You've got some ragged whitespace here, as well as unused semicolons.

PR #877, opened 2024-02-26
Random stuff

moshimeow:


moshimeow:
Pointing out `Add cross-platform skshaderc binary support` ahead of time - it works, but it's kinda ad hoc and we may want to redo the way we load skshaderc binaries in the first place

toby018:
Hello, are you still doing hand tracking? I haven't seen you update in a long time.

PR #909, opened 2024-03-19
Fix: UI.PushEnabled does not visually affect certain elements

ponugotimanasaqc:
Add New window to show UI elements when they are disabled.


ponugotimanasaqc:
Creating a new branch. We will see all the changes in the branch UIPushEnabled2#791

PR #949, opened 2024-04-05
Make `SK.GetStepper` thread safe

maluoi:


StereoKit/Framework/Steppers.cs
```diff
@@ -80,11 +86,14 @@ public void Step()
...
-							_steppers.Add(action.stepper);
+						{
+							lock (_stepperLock) _steppers.Add(action.stepper);
+						}
 						break;
 					case ActionType.Remove:
-						if (_steppers.Remove(action.stepper))
-							action.stepper.Shutdown();
+						bool result = false;
+						lock (_stepperLock) _steppers.Remove(action.stepper);
```
keveleigh, line 95: Doesn't look like `result` is getting set to anything other than `false` here

keveleigh: Whee my first PR haha https://github.com/StereoKit/StereoKit/pull/951

PR #1048, opened 2024-07-17
feat: Added == and != operator overrides to Vec.cs classes

notpatern:


maluoi:
Thanks for the PR! Sign-off is good and changes are done right :) `/master` is intended to be a snapshot of the latest stable version, so this should be a PR against the `/develop` branch.

However! Equality operators for vectors are a bit of a footgun, so I've intentionally avoided them for SK's Vecs. Floating point comparison in particular is the root issue, two floats are "never the same" unless they're straight up copies of eachother! I see you're doing a `Distance`, which is generally the right approach, but it should be a "less than small value/epsilon" rather than "equal to 0". `Distance` is also one of the pricier methods as it involves a `Sqrt`, so a small improvement would be to use `DistanceSq` and a squared epsilon.

But because there's a lot of these small details here, it's kinda dangerous to be doing this in an operator overload, as it's obscuring complex behavior and performance implications. I've seen math libraries that do this before, and they have _always_ had performance problems hiding in sneaky places!

If you want to do this as a static and/or instance method instead, I'd accept that as a PR :)

notpatern:
Alright! Thanks Nick!

I'll reword that PR and post it again later

PR #1054, opened 2024-07-23
Fixes Quest hand menu events

maluoi:
Refactored OpenXR extensions to be aware of their current state: unavailable, active, rejected, or disabled. This allows for a more central store of information about OpenXR extension status.

Added some defensive code in xr view location and swapchain acquisition.


StereoKitC/xr_backends/anchor_openxr_msft.cpp
```diff
@@ -210,7 +210,7 @@ bool32_t anchor_oxr_msft_persist(anchor_t anchor, bool32_t persist) {
 
 bool32_t anchor_oxr_get_perception_anchor(anchor_t anchor, void **perception_spatial_anchor) {
 #if defined(SK_OS_WINDOWS_UWP)
-	if (!xr_ext_available.MSFT_perception_anchor_interop) return false;
+	if (!xr_ext.MSFT_perception_anchor_interop) return false;
```
keveleigh, line 213: Should this be `if (xr_ext.MSFT_perception_anchor_interop != xr_ext_active) return false;`?

maluoi: I did set it up so that this is _technically_ safe and correct, but yeah, I definitely meant to switch 'em all to the more explicit format. Looks like the #ifdefs tripped me up here. Nice catch, thanks!

PR #1086, opened 2024-08-26
Show EGL error information for debugging purposes

Pandapip1:


maluoi:
Heyas, thanks for the PR! Could you instead make this to the [sk_gpu](https://github.com/StereoKit/sk_gpu/tree/master/src) repository? The .h file in this repository is a copy of the amalgamated version of sk_gpu.

Then it can be tagged, and [pulled into /develop](https://github.com/StereoKit/StereoKit/blob/develop/CMakeLists.txt#L210-L213). The latest version of SK no longer directly reference _just_ the .h file.

I think in this case, I'd also probably prefer hex values for the EGL errors instead of extra strings.

PR #1090, opened 2024-08-30
Add overload to Model Ray intersection to return the relevant ModelNode

ponugotimanasaqc:


StereoKitC/stereokit.h
```diff
@@ -1413,6 +1413,7 @@ SK_API void          model_recalculate_bounds_exact(model_t model);
 SK_API void          model_set_bounds              (model_t model, const sk_ref(bounds_t) bounds);
 SK_API bounds_t      model_get_bounds              (model_t model);
 SK_API bool32_t      model_ray_intersect             (model_t model, ray_t model_space_ray, cull_ cull_mode, ray_t* out_pt);
+SK_API bool32_t      model_node_ray_intersect(model_t model, ray_t model_space_ray, cull_ cull_mode, ray_t* out_pt, model_node_id* model_node_id);
```
maluoi, line 1416: I think in this case, it would be a little cleaner to add the new `model_node_id` parameter to `model_ray_intersect` instead of adding a new function? This can have a `sk_default(nullptr)` (with a corresponding null check in the implementation) to prevent this from being a "breaking" change. To indicate these things properly in SK's naming convention here in the public API, that would make the parameter name more like a `out_opt_node_id`.

StereoKitC/asset_types/model.cpp
```diff
@@ -273,25 +273,31 @@ bounds_t model_get_bounds(model_t model) {
...
+	return model_node_ray_intersect(model, model_space_ray, cull_mode, out_pt, &intersected_model_nodeid);
+}
+
+bool32_t model_node_ray_intersect(model_t model, ray_t model_space_ray, cull_ cull_mode, ray_t* out_pt, model_node_id* intersected_model_nodeid) {
 	vec3 bounds_at;
 	if (!bounds_ray_intersect(model->bounds, model_space_ray, &bounds_at))
 		return false;
 
 	float closest = FLT_MAX;
 	*out_pt = {};
```
maluoi, line 286: `out_opt_node_id` would also need to be provided a default value of `-1` here, if not `nullptr`

StereoKit/Math/Ray.cs
```diff
@@ -165,6 +166,16 @@ public bool Intersect(Model model, out Ray modelSpaceAt)
 		public bool Intersect(Model model, Cull cullFaces, out Ray modelSpaceAt)
 			=> NativeAPI.model_ray_intersect(model._inst, this, cullFaces, out modelSpaceAt);
 
+		/// <inheritdoc cref="Intersect(Model, Cull, out Ray)"/>
+		/// <param name="modelNode">ModelNode that intersects with ray.</param>
+		public bool Intersect(Model model, Cull cullFaces, out Ray modelSpaceAt, out ModelNode modelNode)
+		{
+			int modelNodeId = -1;
```
maluoi, line 173: C#'s rule for `out` parameters is that the function _must_ give the `out` parameter _a_ value. While C++ does not have this rule, this is a soft rule I follow when writing C++ code as well. So, this initialization can be skipped and delegated to `mode_ray_intersect`, which allows us to create the variable inline: `model_ray_intersect(model._inst, this, cullFaces, out modelSpaceAt, out int modelNodeId)`

Examples/StereoKitTest/Tests/TestNodes.cs
```diff
@@ -6,6 +6,10 @@ class TestNodes : ITest
 	Pose      _modelPose = Pose.Identity;
 	Model     _model;
 	ModelNode _surfaceNode;
+	ModelNode modelRay2Node;
+	ModelNode modelRay3Node;
+	ModelNode ray2ModelNode;
+	ModelNode ray3ModelNode;
```
maluoi, line 12: C#'s official naming convention has private fields such as these starting with and underscore, so `_modelRay2Node`, etc.

Examples/StereoKitTest/Tests/TestNodes.cs
```diff
@@ -187,7 +197,12 @@ static void RecursiveTraversal(ModelNode node, int depth = 0)
 	}
 	/// :End:
 
-	public void Shutdown() { }
+	public void Shutdown() 
+	{
+		// Verify ModelNode and Ray, Ray and ModelNode intersections.
+		Tests.Test(TestModelNodeRayIntersections);
```
maluoi, line 203: Is there a reason this is being done in `Shutdown` instead of at the end of `Initialize` with the rest of the `Test` calls? If not, it would be good to group those together :)

maluoi (changes requested):
Functionally this looks quite good! I do have some nits with the C public API that may warrant a bit more change.

PR #1091, opened 2024-09-03
oculus/touch_controller add homebutton

tigerhu67:
oculus_touch_controller_profile can not trigger home/system button by right controller , modify this for support.
by the way unity openxr plugin support oculus_touch_controller_profile.

### DIFF
--- a/StereoKitC/xr_backends/openxr_input.cpp
+++ b/StereoKitC/xr_backends/openxr_input.cpp
@@ -324,6 +324,7 @@ bool oxri_init() {
 		_bind_rl(&bind_arr, local.action_stick_xy,    path_stick_xy   );
 		_bind_rl(&bind_arr, local.action_stick_click, path_stick_click);
 		_bind   (&bind_arr, local.action_menu,        path_menu_click.left);
+		_bind	(&bind_arr, local.action_menu,		  path_system_click.right);
 		_bind_rl(&bind_arr, local.action_x1,          {path_x_click.left, path_a_click.right} );
 		_bind_rl(&bind_arr, local.action_x2,          {path_y_click.left, path_b_click.right} );
 		if (xr_ext.EXT_palm_pose == xr_ext_active) _bind_rl(&bind_arr, local.action_palm_pose, path_palm_pose);


maluoi:
Hi, could you explain this in more detail? I'm not very sure what you're trying to do here, it looks like you're trying to map touch_controller's Home button to StereoKit's Menu input? Oculus's controller has a Menu button on the left controller, and a Home button at the same spot on the right, and these serve separate purposes.

tigerhu67:
Hello Nick, thank you for your reply. I was originally a developer of Unity3d, and recently I have been using the SK engine to develop OpenXR overlay applications,

I need to use the home button to trigger functions such as recenter, but I found that there is only a ControlMenuButton in Input without the ControlHomeButton property,

Other methods cannot obtain the pressed state of the home button, such as Input.Controller(Handed.Right). IsHomeJustPressed (actually StereoKit Controller does not have IsHomeJustPressed).

I realized that the ControlMenuButton of the SK engine may have similarities in design with the CommonUsages. menuButton of Unity,

That is, the left-hand controller ControlMenuButton represents the menu button, and the right-hand controller's ControlMenuButton represents the system button. But this doesn't work on SK. I checked the source code of the SK engine,

I found that the underlying layer does not map system_click to the right-handed for ControlMenuButton, and I think this is a issue. Actually, compared to the "Oculus Touch Controller Profile" in the unity openxr plugin

In the mapping, unity mapped the system button to the CommonUsages. menuButton on the right hand (refer to Oculus TouchControlProfile. cs). I hope the SK engine can also implement the home button on Oculus controllers, so

Submit this request.

The following code:
// Menu
                    new ActionConfig()
                    {
                        name = "menu",
                        localizedName = "Menu",
                        type = ActionType.Binary,
                        usages = new List<string>()
                        {
                            "MenuButton"
                        },
                        bindings = new List<ActionBinding>()
                        {
                            new ActionBinding()
                            {
                                interactionPath = menu,
                                interactionProfileName = profile,
                                userPaths = new List<string>() { UserPaths.leftHand }
                            },
                            new ActionBinding()
                            {
                                interactionPath = system,
                                interactionProfileName = profile,
                                userPaths = new List<string>() { UserPaths.rightHand }
                            },
                        }
                    },

https://docs.unity3d.com/Packages/com.unity.xr.openxr@1.3/manual/features/oculustouchcontrollerprofile.html
/input/menu/click	menu (Left Hand Only)	Boolean
/input/system/click	menu (Right Hand Only)	Boolean

PR #1125, opened 2024-11-20
Update ui_layout.cpp

mvvvv:
To avoid segmentation fault when closing the app with steam proton on linux.


mvvvv:
The error that I was finally able to trace on linux thanks to gdb on vscode:
Capture d’écran du 2024-11-20 15-24-48

mvvvv:
when debugging step by step I also got an other segmentation fault about layout->window having a huge negative value.

maluoi:
This change went into PR #1127

PR #1136, opened 2024-12-18
fix: force the prior ARM32 version for UWP CMake builds

austinbhale:
Looks like CMake defaults to the newest SDK version, but the latest Windows Kit doesn't include ARM anymore ☠️

(edit: sorry I forgot DCO! Will remediate that later, or you can close this PR)


maluoi:
Yeah, I ran into this a bit on the build server when the image updated recently, and I believe I fixed it there a slightly different way! But I think this is way better than mine. Nice fix, ty!

nw about the remediation, good to have, but I'm not super picky about it.

PR #1191, opened 2025-04-30
Add UISettings.hseparatorScale and update C# usage and tests

nn33sy:
Add a hseparator_scale setting to control the thickness of horizontal separators in the StereoKit UI.

Main Changes:

- Added the hseparator_scale field to ui_settings_t (C) and UISettings (C#).
- Applied hseparator_scale in the ui_hseparator function to dynamically adjust separator thickness.
- Set a default value (0.4f) if no scale is specified which is the historical factor value for the separator thickness.
- Updated the UI panel to include a slider for adjusting hseparator_scale in DemoUISettings.

### DIFF
--- a/Examples/StereoKitTest/Demos/DemoUISettings.cs
+++ b/Examples/StereoKitTest/Demos/DemoUISettings.cs
@@ -40,8 +40,8 @@ class DemoUISettings : ITest
 		UI.HSlider("sl_depth",    ref activeSettings.depth,    0, 40 * U.mm, 0);
 		UI.Label("Rounding", size); UI.SameLine();
 		UI.HSlider("sl_rounding", ref activeSettings.rounding, 0, 40 * U.mm, 0);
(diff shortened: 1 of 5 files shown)


maluoi:
Oh nice, I like this!

One thing I'd like to see is public names that don't specifically call out "Horizontal" (like hseparator_scale should just be separator_scale), so they have more utility in the future if something like a `UI.VSeparator` is added.

nn33sy:
> Oh nice, I like this!
> 
> One thing I'd like to see is public names that don't specifically call out "Horizontal" (like hseparator_scale should just be separator_scale), so they have more utility in the future if something like a `UI.VSeparator` is added.

Thanks Nick for the quick review ! I changed it ! And yes it would be a good idea a VSeparator !

maluoi:
Awesome, thanks for updating it! :) This should be in tonight's preview build!

PR #1196, opened 2025-05-07
OpenXR Extension Refactor

maluoi:
This reworks all StereoKit's 20+ OpenXR extensions to exist as (mostly) standalone files that plug into an extension system! This should make StereoKit's OpenXR extension implementations easier to add, maintain, and reference.

This also includes a bit of an input overhaul, which now allows for adding new input profiles with much broader shapes. A new input API is added internally, which may eventually make it to the public API. Input poses are now all stored in local space, and converted to world space on-demand. This can help with issues related to updating `Renderer.CameraRoot` and experiencing frame delays for certain poses.

This PR does contain two breaking changes for the C API: `input_head` and `input_eyes` now return copies rather than pointers. @mvvvv

Since this touches a lot of platform specific code, I have done a bit of extra testing on devices, but I'd keep an eye out for any XR ext or input related regressions!


StereoKitC/xr_backends/extensions/hand_tracking.cpp~RF1c29c2ca.TMP
```diff
@@ -0,0 +1,123 @@
+/* SPDX-License-Identifier: MIT */
```
keveleigh, line 1: The filename here looks like a temp file was included 🤔 it looks to have different content than the non-temp version, so worth double checking which one is "right"!

PR #1226, opened 2025-06-24
Fix a few warnings and add a missing struct type for render models

keveleigh:
* Fix warning C4018: '<': signed/unsigned mismatch
* Fix warning C4091: 'typedef ': ignored on left of 'sk::xr_render_model_t' when no variable is declared
* FIx missing struct type

### DIFF
--- a/StereoKitC/xr_backends/extensions/ext_render_model.h
+++ b/StereoKitC/xr_backends/extensions/ext_render_model.h
@@ -11,13 +11,13 @@
 namespace sk {
 
 typedef struct xr_render_model_t {
-	XrRenderModelIdEXT id;
-	XrRenderModelEXT   render_model;
-	model_t            model;
-	XrSpace            space;
-	model_node_id*     anim_nodes;
-	int32_t            anim_node_count;
-	XrRenderModelStateEXT state_query;
+	XrRenderModelIdEXT    id;
+	XrRenderModelEXT      render_model;
+	model_t               model;
+	XrSpace               space;
+	model_node_id*        anim_nodes;
+	int32_t               anim_node_count;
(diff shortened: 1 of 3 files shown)


StereoKitC/xr_backends/extensions/ext_render_model.h
```diff
@@ -11,14 +11,14 @@
...
-	int32_t            anim_node_count;
-	XrRenderModelStateEXT state_query;
-};
+	XrRenderModelIdEXT    id;
+	XrRenderModelEXT      render_model;
+	model_t               model;
+	XrSpace               space;
+	model_node_id*        anim_nodes;
+	int32_t               anim_node_count;
+	XrRenderModelStateEXT state_query = { XR_TYPE_RENDER_MODEL_STATE_EXT };
```
maluoi, line 20: This violates some of the "Plain old Data" design I try to stick to in StereoKit, and I believe introduces an invisible constructor? I'd like to avoid this!

keveleigh: Ahh okay! We'll want to change it here as well, which is where I was "inspired" 😄  

https://github.com/StereoKit/StereoKit/blob/b4a113fffd0693acb772b931add2f682c5419656/StereoKitC/xr_backends/extensions/hand_mesh.cpp#L27

maluoi: Whoops! Yeah, that one looks like it was probably an evolution from raw globals to a state struct! That shouldn't have survived 😅 Initialization like that can lead to issues when "restarting" StereoKit in the same process, since relying on global initialization means the variables don't always get reset properly!

keveleigh: Updated in https://github.com/StereoKit/StereoKit/pull/1227

PR #1305, opened 2026-02-03
Add Android APK build/run targets for StereoKitCTest and a flag that helps with interactive debugging

austinbhale:
Based on the native templates, this adds APK build and run targets for `StereoKitCTest` on Android, as well as a property for debugging APKs in Android Studio.

This setup favors editing and testing StereoKit natively rather than consuming it as a dependency in the templates, since I think it's useful for testing/implementing new StereoKit features natively, and mixed debugging with .NET and C++ on VS for Android is out of support + unusable (e.g, segfaults galore). Just like in the templates, you can run it after building the APK.

For Android Studio iterations, the current workflow is pretty decent (most of the following setup is done just once):
1) Disable `SK_STRIP_DEBUG_SYMBOLS` and build the APK target. This ensures debug symbols remain embedded in the `.so`, which Android Studio can locate
2) (repeatable) Open Android Studio > Profile or Debug an APK
3) Click the StereoKit symbols > Ensure the path mappings are configured to show native code
4) File > Project Structure > Modules > Dependencies > Set Module SDK
5) Edit your configuration if needed to run
6) (repeatable) Debug, or a quick workaround is to Run then Attach Debugger (Use default settings)

The samples work well, just the Aliasing demo crashes when changing the settings on Quest.


maluoi:
I'm actually in the process of integrating [`sk_app`](https://github.com/StereoKit/sk_app), which has a number of nice tools for building to Android, among other things! I expect this PR will probably conflict, so I may skip on merging this one.

It does seem like you're doing some nice stuff with debug settings though, so I'll definitely reference what you've added here and make sure sk_app is doing similar!

austinbhale:
Ohh very cool, I'll check that out! Glad you have newer updates there. No worries about merging this in, was a bit of a side quest for wanting better debugging when working on spatial entities. I'll double check if Android Studio can find the symbols again but I think that property would be useful to have upstream (edit: confirmed)

Issue #1287, opened 2025-12-15
String buffer size not correctly handled in sk::ui_input()


guillaume-roche:
## Description

When I add N more characters than allowed to my string in the input field, I have to hit N times backspace before it starts removing characters from the end of the string.

## Platform / Environment

- Android 12 / SDK32.
- Lynx R-1
- SK v0.3.x (but I believe the issue is also on the `develop` branch).

## Logs or exception details

No particular logs or exception, since the bug is only behavioral. However I think I have found the bug to be there (stereokit_ui.cpp)

```
utf_insert_char(buffer, buffer_size, utf_advance_chars(buffer, skui_input_carat), add);
skui_input_carat += 1;
skui_input_carat_end = skui_input_carat;
result = true;
```

Explanation:

`utf_insert_char()` returns false when the buffer is full, and no extra character is appended to the string. However the return value of that function is not checked, and `skui_input_carat` is always incremented.

I think a fix would be something like:

```
if (utf_insert_char(buffer, buffer_size, utf_advance_chars(buffer, skui_input_carat), add)) {
    skui_input_carat += 1;
    skui_input_carat_end = skui_input_carat;
    result = true;
}
```

guillaume-roche:
Fixed by https://github.com/StereoKit/StereoKit/pull/1306

PR #1306, opened 2026-02-05
Fix sk::ui_input buffer size handling

guillaume-roche:
When I add N more characters than allowed to my string in the input field, I have to hit N times backspace before it starts removing characters from the end of the string.

This patch fixes this issue by handling the return value of utf_insert_char().

Tested for my use case, it appears to work well, however maybe a more in-depth testing is required?

Fixes https://github.com/StereoKit/StereoKit/issues/1287.


maluoi:
Sorry, I seem to have missed when you opened your initial issue! This fix looks great to me!

PR #1308, opened 2026-02-08
Add XR_META_environment_depth extension support

austinbhale:
Implements Meta Quest environment depth sensing via [`XR_META_environment_depth` OpenXR extension](https://developers.meta.com/horizon/documentation/native/android/mobile-depth/). Verified on Quest 3/3S.

Given the hard requirement of acquiring depth maps between `xrBeginFrame`/`xrEndFrame`, I figured this would better align with StereoKit's statelessness over a render event callback.

#### Demos
- [Point cloud visualization + controls](https://drive.google.com/file/d/1cSy11k7U23hNU0Hrf_y2OJbo9Zc8DT7K/view?usp=sharing)
- [Virtual object (hard) occlusion](https://drive.google.com/file/d/1oAFTb945Lhvyme0bK6i8kfWy2XtYxmPR/view?usp=sharing)

Would be cool to see a soft occlusion impl in the future!


maluoi:
Love this, looks really clean! Here's some notes.

## XR_ANDROID_depth_texture

Related extension, not in the official spec yet, but documented on the Android docs:
https://developer.android.com/develop/xr/openxr/extensions/XR_ANDROID_depth_texture

I haven't read it in detail yet, and don't feel like you need to implement this. It looks like it maybe more directly sensor data focused, where the META one is focused on enabling the occlusion experience? Ideally it would be great to fit these both under the same public API, but I'm not sure if they'll do it nicely.

## Existing SK Features
Currently existing are [`World.OcclusionEnabled`](https://stereokit.net/Pages/StereoKit/World/OcclusionEnabled.html), [`World.OcclusionMaterial`](https://stereokit.net/Pages/StereoKit/World/OcclusionMaterial.html) and [`SystemInfo.worldOcclusionPresent`](https://stereokit.net/Pages/StereoKit/SystemInfo/worldOcclusionPresent.html). These are currently backed by Microsoft's Scene Understanding world mesh, where `OcclusionEnabled` turns on the meshing feature, and the mesh chunks get drawn to the scene with the provided `OcclusionMaterial`.

Depth textures look like they would map pretty well to `OcclusionEnabled`, but may not exactly make sense with `OcclusionMaterial` as is. Maybe something like a `enum OcclusionSource {None, Mesh, Depth}` to clarify if the Material is applied to a Mesh, or is responsible for writing the depth (like with your "env_depth_occlusion.hlsl")? Alternatively (and I might prefer it), we could deprecate `OcclusionMaterial` and no-op it from this data source. Materials are still accessible and modifiable via `Material.Find` and a proper `Default` asset id for those that really want to dig.

It would be nice to mix hand occlusion in here, so something needs added for that. Maybe a `World.OcculusionFlags` property with a `OcclusionFlag` bitfield enum that includes `Hands`, and a `World.OcclusionCapabilities(OcclusionFlag)` to query system ability? In the future it might distinguish between humans and environment, so keeping it extendable seems wise.

## Direct API

The `World.Occlusion` APIs obscure direct access to the underlying ext, but surfaces the core feature of `XR_META_environment_depth` pretty well. We could drop EnvironmentDepth.cs entirely in favor of the `World.Occlusion` route, but you would miss out on some of the additional fun things you're doing with the depth data!

To continue down this route, I think I'd want to lump it in with the more general concept of sensor data? Perhaps `EnvironmentDepth` becomes a `Sensor` class that might host depth camera data (XR_META_environment_depth, XR_ANDROID_depth_texture) and color camera data (XR_OCULUS_external_camera, other future passthrough texture stuff)? This class wouldn't need to be fully implemented, just the depth part, but it might want to consider if and how other bits might co-exist. Like your `EnvironmentDepthView` would be more like a `SensorDataPose` or something.

austinbhale:
Thanks for the great suggestions! I enjoyed seeing how depth-based world occlusion turned out. After iterating some more, I've reworked things based on your feedback. Here's where it landed:

### Depth vendor abstraction like `XR_ANDROID_depth_texture`

Added an intermediary `sensor_depth` layer that dispatches to backends via a `sensor_depth_system_` enum (currently just `openxr_meta`). Adding an `openxr_android` backend should slot in cleanly.

### Occlusion

Went with a flags-based approach. New `occlusion_caps_` bit field:

- Mesh — Scene Understanding (HoloLens)
- Depth — depth texture occlusion (META environment depth)
- Hands — when combined with Depth, hands occlude virtual content (without it, hand removal is enabled)

`World.Occlusion` (get/set) and `World.OcclusionCapabilities` replace the prior bool way. `OcclusionEnabled` and `OcclusionMaterial` are deprecated though they could be obsolete since HoloLens already is? For the depth path, `world.cpp` manages the sensor lifecycle automatically and a built-in depth pre-pass shader writes environment depth into the z-buffer.

### Sensor API

Renamed to `Sensor.Depth` nested under a `Sensor` class that can host future data sources (color camera, etc.). `EnvironmentDepthView` became `SensorDepthView`, and the frame struct uses `views[2]` instead of named left/right. Also changed the depth texture from vertically-packed to a 2D array texture with 2 layers. The API surface:

- Start(SensorDepthCaps) / Stop() / SetCapabilities()
- Texture — direct GPU texture access
- TryGetLatestFrame() — per-frame metadata (poses, FOV, near/far, display + capture timestamps)
- TryGetLatestData<T>() — async CPU readback with matching metadata

Added `capture_time` alongside `display_time` for v2 timestamp support. The CPU readback in `sensor.cpp` uses a double-buffered pipeline - a pending readback runs asynchronously while the previous completed result stays available, so callers never stall the GPU. Frame metadata is stashed at submission time, so the CPU data is always paired with the correct poses. Data is typically 1-2 frames behind the GPU texture. Although a bit premature, the readback layer is reusable by future sensor types.

`World.Occlusion` remains as a simple API, and `Sensor.Depth` is there for direct access (point clouds, spatial mapping, etc.). Thoughts on the current shape here?

StereoKitC/xr_backends/extensions/meta_environment_depth.cpp
```diff
@@ -0,0 +1,409 @@
...
+	XrEnvironmentDepthProviderCreateInfoMETA provider_info = { XR_TYPE_ENVIRONMENT_DEPTH_PROVIDER_CREATE_INFO_META };
+	result = xrCreateEnvironmentDepthProviderMETA(xr_session, &provider_info, &local.provider);
+	if (XR_FAILED(result)) {
+		log_warnf("XR_META_environment_depth: xrCreateEnvironmentDepthProviderMETA failed: [%s]", openxr_string(result));
+		xr_ext_meta_environment_depth_destroy();
+		return xr_system_fail;
+	}
+
+	XrEnvironmentDepthSwapchainCreateInfoMETA swapchain_info = { XR_TYPE_ENVIRONMENT_DEPTH_SWAPCHAIN_CREATE_INFO_META };
+	result = xrCreateEnvironmentDepthSwapchainMETA(local.provider, &swapchain_info, &local.swapchain);
```
maluoi: This is creating swapchain during extension initialization! This means we may end up allocating swapchain memory regardless of whether we use `xr_ext_meta_environment_depth_start` or not. We should probably lazy allocate a lot of stuff on `xr_ext_meta_environment_depth_start`, and _only_ do base ext and property setup here in initialize.

StereoKitC/shaders_builtin/shader_builtin_sensor_depth_prepass.hlsl
```diff
@@ -0,0 +1,88 @@
...
+	float4 near_vs = mul(float4(ndc.x, -ndc.y, 0, 1), sk_proj_inv[input.view_id]);
+	near_vs /= near_vs.w;
+	float3 ray_dir_vs = normalize(near_vs.xyz);
+
+	// Rotate view-space ray to world space
+	float3x3 view_rot = float3x3(
+		sk_view[input.view_id][0].xyz,
+		sk_view[input.view_id][1].xyz,
+		sk_view[input.view_id][2].xyz);
+	float3 ray_dir_ws = normalize(mul(ray_dir_vs, transpose(view_rot)));
```
maluoi: Most of this code can, and probably should be shifted to the vertex shader! Since this is a full-screen effect, any cost you pay in the fragment shader really get multiplied!

VS gets called 8-12 times per frame
PS gets called 1680x1760(resolution)x~4(msaa)x2(eyes) times per frame

StereoKitC/shaders_builtin/shader_builtin_sensor_depth_prepass.hlsl
```diff
@@ -0,0 +1,88 @@
+#include "stereokit.hlsli"
+//--name = default/shader_depth_prepass
```
maluoi: For builtins, I've been using `sk/` for the names rather than `default/`.

maluoi (changes requested):
Sorry for being slow, this looks a lot better to me!

I do have some perf/resource concerns here that I've noted inline.

StereoKitC/shaders_builtin/shader_builtin_sensor_depth_prepass.hlsl
austinbhale: Updated to `sk/depth_prepass`

StereoKitC/shaders_builtin/shader_builtin_sensor_depth_prepass.hlsl
austinbhale: Good call - tried to move all parts that don't need to be per-pixel to the vertex shader

StereoKitC/xr_backends/extensions/meta_environment_depth.cpp
austinbhale: Done! The provider and swapchain are now deferred to Start

austinbhale:
Glad it's looking better and thanks for the suggestions! Let me know if you have any more or want to refactor the existing ones - I'm going to take another pass at the code. Also have been testing the CPU readback path separately from the demo

maluoi:
Thanks for the changes! I think this looks good enough for merging now :)

PR #1320, opened 2026-02-20
Add Android mic input preset selection support

austinbhale:
More info about each preset can be found [here](https://developer.android.com/reference/android/media/MediaRecorder.AudioSource) and an example of recording/playback disparities in the Mic Recording demo. There should be no effect on non-Android platforms.

### DIFF
--- a/Examples/StereoKitTest/Demos/DemoRecordMic.cs
+++ b/Examples/StereoKitTest/Demos/DemoRecordMic.cs
@@ -3,6 +3,7 @@
 // Copyright (c) 2019-2023 Nick Klingensmith
 // Copyright (c) 2023 Qualcomm Technologies, Inc.
 
+using System;
 using System.Collections.Generic;
 using StereoKit;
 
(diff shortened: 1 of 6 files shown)


maluoi:
Can you add an enum for the preset instead of an Android specific integer? It looks like this concept may have equivalents on Windows and MacOS! Don't worry about implementing them, but making an enum that can leave room for the other platforms in the future might be nice.

On Windows, I see a [AUDIO_SIGNALPROCESSINGMODE_](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/ksproperty-audiosignalprocessing-modes) and [maybe here](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/audio-signal-processing-modes).

On MacOS, [this one](https://developer.apple.com/documentation/avfaudio/avaudiosession/category-swift.struct/record)?

austinbhale:
Good call! Replaced the int with a `mic_input_preset_` enum. The values could map naturally on other platforms in the future - e.g. Unprocessed == [Windows](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/audio-signal-processing-modes) Raw || [Apple](https://developer.apple.com/documentation/avfaudio/avaudiosession/mode-swift.struct/measurement) `.measurement`, VoiceCommunication == Windows Communications || Apple `.voiceChat`, etc. Also added the older OpenSL preset support alongside AAudio

Issue #1303, opened 2026-02-03
Crash due to permission requests on Android


guillaume-roche:
## Description

I tried to integrate the `develop` branch in our app to anticipate any issue that would potentially occur with SK next release.

My app actually has an Android Service (not an activity), that draws some UI in an overlay, above XR apps. As soon as `sk_init` is called, my service crashes. From my investigation, it seems that SK assumes running in an activity, and calls the `requestPermissions()` method that doesn't exist in the `Service` class.

To give a bit more Context (pun intended)
- In order to initialize SK, I set my Service Context into `settings.android_activity`.
- My service is signed with the Android platform keys, so it doesn't need to dynamically requests permissions, it will work as long as they are correctly declared in the manifest.

Everything is fine with SK v0.3.x.

## Platform / Environment

- Device: Lynx R-1.
- OS: Android arm64.
- OpenXR: Monado based runtime.

## Logs or exception details

Crash error message:
```
JNI DETECTED ERROR IN APPLICATION: JNI GetMethodID called with pending exception java.lang.NoSuchMethodError: no non-static method "Lcom/lynx/ui/service/LynxUIService;.requestPermissions([Ljava/lang/String;I)V"
```

Stack trace:
```
[libc.so] abort 0x0000007799315988
[libart.so] art::Runtime::Abort(const char *) 0x000000746a635214
<unknown> 0x0000007788a16bc8
<unknown> 0x0000007788a16188
[libart.so] art::JavaVMExt::JniAbort(const char *, const char *) 0x000000746a458988
[libart.so] art::JavaVMExt::JniAbortV(const char *, const char *, std::__va_list) 0x000000746a458a04
[libart.so] art::(anonymous namespace)::ScopedCheck::AbortF(char const*, ...) 0x000000746a44a96c
[libart.so] art::ScopedCheck::CheckPossibleHeapValue(art::ScopedObjectAccess &, char, art::JniValueType) 0x000000746a449060
[libart.so] art::ScopedCheck::Check(art::ScopedObjectAccess &, bool, const char *, art::JniValueType *) 0x000000746a447bac
[libart.so] art::CheckJNI::GetMethodIDInternal(const char *, _JNIEnv *, _jclass *, const char *, const char *, bool) 0x000000746a44be88
[libStereoKitC.so] sk::_JNIEnv::GetMethodID(sk::_jclass *, const char *, const char *) 0x00000073e37e3ddc
[...]

PR #1321, opened 2026-02-26
Platform upgrades

maluoi:
This is a complete re-write of StereoKit's platform abstraction! This moves from an in-situ implementation to [a separate repository](https://github.com/StereoKit/sk_app), and was built from the ground up to be cross platform on Win32, Linux, Android, and MacOS.

This does include breaking changes for shaders again! The layer select strategy I was carrying forward from D3D11 for back-compat was not working on certain XR headsets due to a missing layer select extension. However, VK_KHR_multiview is required on all Vulkan 1.1 devices for a minimum of 6 views, does have a small perf benefit over layer select, and behaves _much_ better in Vulkan than it did in GLES. I'm pretty happy with VK_KHR_multiview now, and it does actually simplify layer/instance shader code as well.

This is a good bit of new code, so please let me know if you see any bugs! Keep your eyes out for anything app lifetime related.

Notable new features:
- MacOS support for the Simulator! C# and C/C++! (OpenXR might work if the runtime supports Vulkan, not tested)
- Native file picker on Android, which handles file permissions!
- Cmake improvements for Android builds.
- Simulator on Linux restores window position.
- Switched to VK_KHR_multiview for single-pass strategy. Some perf benefits.

## For porting shaders, see this for reference:

```diff
#include "stereokit.hlsli"

//--name = sk/unlit

float4       color     = {1,1,1,1};
float4       tex_trans = {0,0,1,1};

//--diffuse = white
Texture2D    diffuse   : register(t0);
SamplerState diffuse_s : register(s0);

struct vsIn {
	float4 pos  : SV_Position;
	float3 norm : NORMAL0;
	float2 uv   : TEXCOORD0;
	float4 col  : COLOR0;
};
struct psIn {
	float4 pos   : SV_POSITION;
	float2 uv    : TEXCOORD0;
	half4  color : COLOR0;
-	uint view_id : SV_RenderTargetArrayIndex;
};

- psIn vs(vsIn input, uint id : SV_InstanceID) {
+ psIn vs(vsIn input, sk_ids_t ids) {
	psIn o;
-	o.view_id = id % sk_view_count;
-	id        = id / sk_view_count;
	
-	float4 world = mul(float4(input.pos.xyz, 1), sk_inst    [id].world);
-	o.pos        = mul(world,                    sk_viewproj[o.view_id]);
+	float4 world = mul(float4(input.pos.xyz, 1), sk_inst    [ids.inst].world);
+	o.pos        = mul(world,                    sk_viewproj[ids.view]);

	o.uv    = (input.uv * tex_trans.zw) + tex_trans.xy;
-	o.color = input.col * color * sk_inst[id].color;
+	o.color = input.col * color * sk_inst[ids.inst].color;
	return o;
}

half4 ps(psIn input) : SV_TARGET {
	return diffuse.Sample(diffuse_s, input.uv) * input.color;
}
```

Instead of using `skr_ids_t`, you can also just add these values explicitly. There's no magic, just some built-ins.

```diff
- psIn vs(vsIn input, uint id : SV_InstanceID) {
+ psIn vs(vsIn input, uint id : SV_InstanceID, uint view : SV_ViewID) {
```


guillaume-roche:
Hi Nick!

Tried it briefly on our service that renders a SK UI in an OpenXR overlay. I get the following error:

```
00:10:54.751  4096-4209  SK_Lynx_UI_Service    D  Initializing StereoKit v0.4.0 Android ARM64...
00:10:54.751  4096-4209  SK_Lynx_UI_Service    D  Initializing Assets
00:10:54.778  4096-4209  SK_Lynx_UI_Service    D  Initializing Platform
00:10:54.779  4096-4209  lynx.ui.servic        A  java_vm_ext.cc:579] JNI DETECTED ERROR IN APPLICATION: JNI FindClass called with pending exception java.lang.NoSuchMethodError: no non-static method "Lcom/lynx/ui/service/LynxUIService;.getWindow()Landroid/view/Window;"
                                                  java_vm_ext.cc:579]   at void com.lynx.ui.service.LynxUIService.runService() (LynxUIService.java:-2)
                                                  java_vm_ext.cc:579]   at void com.lynx.ui.service.LynxUIService.$r8$lambda$nP-eMrsoVzmmAzN4duNRQ1LnGaI(com.lynx.ui.service.LynxUIService) (LynxUIService.java:-1)
                                                  java_vm_ext.cc:579]   at void com.lynx.ui.service.LynxUIService$$ExternalSyntheticLambda0.run() (D8$$SyntheticClass:0)
                                                  java_vm_ext.cc:579]   at void java.lang.Thread.run() (Thread.java:920)
                                                  java_vm_ext.cc:579] 
                                                  java_vm_ext.cc:579]     in call to FindClass
                                                  java_vm_ext.cc:579]     from void com.lynx.ui.service.LynxUIService.runService()
```

I guess this is a similar issue as https://github.com/StereoKit/StereoKit/issues/1303. Since I run a service, there is no `getWindow()` method.

maluoi:
Thanks for trying this out! I think the service situation was the bit I was most worried about. I guess it should be worth it for me to find some time to figure out and add some Service based tests so this doesn't keep biting me!

guillaume-roche:
Tested again.

My service no longer crashes! However, I'm experiencing serious rendering issues. I have a test UI that consist of a simple window with buttons, but almost none of it is displayed. I just see what seems to be a button, without any text inside. I don't see the window, nor its background, 3d hands aren't displayed either...

Rendering seems fine on the `develop` branch.

I didn't notice any error that could explain this regression, but if I can look something in particular to help diagnosing it, let me know! My platform is Android aarch64 with monado based OpenXR runtime.

maluoi:
@guillaume-roche Do you have any SK logs for that, or a screenshot you might be able to share? I haven't seen this myself, and I'm not super sure what this might be.

It vaguely sounds like maybe depth related issues? Like if the skybox was drawing overtop of opaque content without testing or something?

I'll poke around and see if anything jumps out in the code.

guillaume-roche:
I was in AR mode (video passthrough) so, there is no skybox in my tests.

Screenshot on the develop branch;
image

Screenshot on this branch (it only renders on the left screen by the way):
image

I don't see any other UI than the black button and the check box.

guillaume-roche:
Hi again,

I did more testing on my side. ~~Oddly enough, our service that displays a SK UI in an OpenXR overlay displays correctly! The SK configuration is similar between the service and the apps. I will continue investigating.~~ 
**Update**: after a full rebuild of the service, I observe the same rendering issues.

Attaching a logcat, if you want to have a look. You will see both the "Dev App" and the service starting.
[lynx-system.log](https://github.com/user-attachments/files/25711784/lynx-system.log)

maluoi:
Ohh thanks for the logs! Are you sure this worked on the `/develop` branch? After Vulkan got merged in?

I see this one in there, which does concern me a fair bit!
> [sk_renderer] Device extension 'VK_EXT_shader_viewport_index_layer' not available, multi-view rendering will not work

I rely on this extension pretty heavily for rendering at the moment. I thought this one was more common than it seems to actually be. It's not specific to the `/feature/sk_app` branch though, this should be on any build since the switch to Vulkan!

In testing, my Vive XR Elite seems to be missing the same extension. It crashes instead of doing render artifacts like these, but could be just a different reaction to the same problem! I'll be working on a rendering method for devices without `VK_EXT_shader_viewport_index_layer`, but if the problem for you is still specific to `/feature/sk_app` and not latest `/develop`, that could indicate some other issue! If you could verify that's the case, I'd appreciate it!

guillaume-roche:
So, I did test again on the latest `develop` branch, with full cleanup of previous builds of my project and the stereokit lib.

I confirm that the display is working fine in that case, although I still have the Vulkan warning in the logs.

maluoi:
@guillaume-roche Latest change reworks the renderer to avoid `VK_EXT_shader_viewport_index_layer` entirely.  There's some other Vulkan init fixes in here, which may also improve the situation, and I also see a perf improvement vs. the previous layer select method. It will require a full re-build (x64 then arm64) due to shader and shader compiler changes!

This fixes Vive XR Elite for me, which was running an older GPU driver, so this may have a positive impact here? But since `/develop` was still working for you, I'm less certain if this is what you're running into. If a Debug build doesn't work, try a Release build too. Vive's older GPU driver seems to not like debug shader code a whole lot.

I would love to know if this improves things for you at all!

guillaume-roche:
Hi Nick,

I replied to you in private on Discord, if you want to check.

Cheers,

PR #1335, opened 2026-03-30
fix: add CMAKE_TOOLCHAIN_FILE pointing to the NDK's android.toolchain

austinbhale:
When building `develop`, I've been running into an error with CMake's built-in Android support because the `platforms/` directory was removed:

```
 Android: The system root directory needed for the selected Android version and architecture does not exist:
      .../ndk/29.x.x/platforms/android-32/arch-arm64
```

Using the NDK's toolchain file fixes this and is recommended [here](https://developer.android.com/ndk/guides/cmake#toolchain_file)

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -322,6 +322,7 @@
       "generator":   "Ninja",
       "cacheVariables": {
         "CMAKE_SYSTEM_NAME":    "Android",
+        "CMAKE_TOOLCHAIN_FILE": "$env{NDK}/build/cmake/android.toolchain.cmake",
         "CMAKE_ANDROID_NDK":    "$env{NDK}",
         "ANDROID_PLATFORM":     "android-32",
         "CMAKE_SYSTEM_VERSION": "32",


maluoi:
What version of CMake are you using here? I'm currently on 3.31.6, and have seen some recommendations for CMake 3.21+ related to building for Android?

The docs cited there look pretty old (cmake 3.6?), and likely are not really relevant anymore, I've settled on StereoKit using CMake's builtin support for Android rather than using NDK's toolchain. Switching to the NDK toolchain instead of the builtin functionality would require more work than just this!

If you're on an older CMake, the more correct action could just be requiring CMake 3.21?

austinbhale:
I'm on CMake 4.1 with an ARM64 Windows machine, and I believe that's the latest docs.

I think the issue is still unfixed on CMake's [master](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Android-Determine.cmake#L291-296). They only check for AMD64 when setting the NDK host tag on Windows. Since the NDK toolchain would be extra work here, I tried more targeted fixes - e.g., `CMAKE_ANDROID_NDK_TOOLCHAIN_HOST_TAG` gets overwritten and `CMAKE_SYSTEM_VERSION` doesn't get configured.

I'll look into filing a CMake bug!

PR #1374, opened 2026-06-24
Adds the Android trackables spatial anchor backend

austinbhale:
Adds an Android spatial anchor backend (`anchor_system_openxr_android`) built on `XR_ANDROID_trackables` (`xrCreateAnchorSpaceANDROID`). It mirrors the existing `msft_anchors` backend almost 1:1, so StereoKit's `Anchor` API now works on Android XR.

## Why this extension

`xrCreateAnchorSpaceANDROID` is the structural twin of MSFT's `xrCreateSpatialAnchorMSFT` + `xrCreateSpatialAnchorSpaceMSFT`: create a free-pose anchor (`trackable = XR_NULL_TRACKABLE_ANDROID`), get an `XrSpace`, and locate it each frame. It slots straight into the `anchor_system_` switch. The cross-vendor `XR_EXT_spatial_entity` path (upstream `feature/spatial_entities`) is a larger, unmerged rewrite.

## Scope
`android_anchors.cpp/.h` (new), `anchor.cpp/.h`, `_registration.cpp`, and a `CMakeLists.txt` bump of the OpenXR loader **1.1.49 → 1.1.60** (the trackables symbols are absent before 1.1.57).

## Limitations
Free anchors have no persistence (`XR_ANDROID_device_anchor_persistence` is separate), so `Capabilities` reports `Stability` only; `persist`/`clear_stored` are no-ops - matching the MSFT backend's no-persistence branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,7 +209,7 @@ include("cmake/CPM.cmake")
 #### OpenXR - https://www.khronos.org/openxr/ ####
 if (SK_BUILD_OPENXR_LOADER)
   CPMAddPackage( # We're scraping the comment after the NAME for some other scripts
-    NAME openxr_loader # 1.1.49
+    NAME openxr_loader # 1.1.60
     GITHUB_REPOSITORY KhronosGroup/OpenXR-SDK
     GIT_TAG release-1.1.60
     GIT_SHALLOW 1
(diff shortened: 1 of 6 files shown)


maluoi:
FYI, for the copyright bit, you should probably use your own name there 😅

This extension is actually listed as deprecated already. I like that it fits into Anchor as a backend, but I'm hesitant about hopping onto an extension that never quite got off the ground.

austinbhale:
Aw man well that was a short-lived extension! I think I’ll just use it till spatial entities are in - when do you expect that to be?

Also G2K about the copyright header moving forward.

maluoi:
I'll see about pushing entities forward this week! If nothing else, I'll try and get the branch into a good state that's useful for review.

 I _do_ like the idea of making an Anchor backend based on it, so I might do that in addition to a pure entities API.

austinbhale:
Oh sweet, looking forward to it. Let me know if you'd like any input on that. Would enjoy using the same anchor API (hopefully with a storable capability this time) but no worries if it's better suited as a replacement

PR #1386, opened 2026-07-13
Add AHardwareBuffer texture import

austinbhale:
Adds `Tex.FromHardwareBuffer` / `tex_create_from_hardware_buffer`, which imports an `AHardwareBuffer` as an external texture with no copy (YCbCr camera and decoder buffers come in via sampler conversion), so camera passthrough frames and hardware-decoder output go straight to the GPU. It rides on sk_renderer's existing `skr_tex_create_external_ahb` and is additive and Android-only. The C API returns null on a null buffer or missing device capability, and an error-state tex per the usual asset convention when the import fails; the C# wrapper resolves all of these to null. `Tex.GetHardwareBuffer` returns the backing buffer.

### DIFF
--- a/StereoKit/Assets/Tex.cs
+++ b/StereoKit/Assets/Tex.cs
@@ -491,6 +491,41 @@ namespace StereoKit
 		public IntPtr GetNativeSurface()
 			=> NativeAPI.tex_get_surface(_inst);
 
+		/// <summary>Imports an Android AHardwareBuffer as an external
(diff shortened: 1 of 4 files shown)


maluoi:
For this one, I think I'm good with the Tex additions! But `render_get_device` is a leftover from SK's D3D only days. Vulkan objects should be surfaced through a `Backend.Vulkan` instead, which doesn't exist yet.

austinbhale:
Updated! Looking forward to `Backend.Vulkan` 👽

PR #1411, opened 2026-08-28
Add test running steps to CI

keveleigh:
Found a test issue with the Windows build locally, so figured it could be useful to run these in CI. Testing the feasibility here.

### DIFF
--- a/.github/workflows/verify_compile.yml
+++ b/.github/workflows/verify_compile.yml
@@ -25,8 +25,18 @@ jobs:
 
     - name: Build Native Win32 x64
       run: |
-        cmake --preset Win32_x64_Release_Fast
+        cmake --preset Win32_x64_Release_Fast -DSK_BUILD_TESTS=ON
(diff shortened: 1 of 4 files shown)


maluoi:
This branch may have some interesting bits in it related to this? It's a little stale, but was pretty close. This is before the SKTest project got added though.

https://github.com/StereoKit/StereoKit/tree/feature/gha_tests

maluoi:
The `skr_future_wait(&future);` over in `compute_buffer.cpp` looks kinda scary, that's a 3 frame drop any time it's called! What's it supposed to solve?

maluoi:
Also, `Win32_x64_Release_Fast` is different from `dotnet run -c Release`, the `dotnet` command here will build the `Win32_x64_Release` preset instead, which doesn't use the existing build from `Win32_x64_Release_Fast`, and instead recompiles from scratch using the glslang backend of skshaderc, which is why that build takes so long.

It might be fine to just use Debug builds here?

